PR #3, opened 2023-03-29
Use a concrete type for functions instead of Rust closures

mikeday:
Minor side benefit: Format can now automatically derive Debug!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -29,6 +29,14 @@ enum Expr {
     Record(Vec<(String, Expr)>),
 }
 
+#[derive(Clone)]
+enum Func {
+    Expr(Expr),
+    Fst,
+    Snd,
+    U16Be,
+}
+
 /// Binary format descriptions
 ///
 /// A subset of binary formats can be modelled as [regular expressions]:
@@ -74,7 +82,7 @@ enum Format {
     /// Restrict a format to a sub-stream of a given number of bytes
     Slice(Expr, Box<Format>),
     /// Transform a decoded value with a function
-    Map(fn(&Value) -> Value, Box<Format>), // TODO: Decouple from `Value`
+    Map(Func, Box<Format>),
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -134,6 +143,36 @@ impl Expr {
...
+                },
+                _ => panic!("U16Be: expected (_, _)"),
+            },
+            Func::Stream => match arg {
+                Value::Seq(vs) => {
+                    // FIXME could also condense nested sequences
+                    Value::Seq(vs.into_iter().filter(|v| *v != Value::Unit).collect())
+                }
+                _ => panic!("Stream: expected Seq"),
+            },
```
brendanzab, line 171: Mildly terrified about what type this would have 😅

src/main.rs
```diff
@@ -134,6 +143,36 @@ impl Expr {
...
+                Value::Pair(fst, snd) => match (fst.as_ref(), snd.as_ref()) {
+                    (Value::U8(hi), Value::U8(lo)) => Value::U16(u16::from_be_bytes([*hi, *lo])),
+                    (_, _) => panic!("expected (U8, U8)"),
+                },
+                _ => panic!("U16Be: expected (_, _)"),
+            },
+            Func::Stream => match arg {
+                Value::Seq(vs) => {
+                    // FIXME could also condense nested sequences
+                    Value::Seq(vs.into_iter().filter(|v| *v != Value::Unit).collect())
```
brendanzab, line 168: Note: Could use the `matches!` macro if you wanted to avoid needing `PartialEq`, but not a big deal.

mikeday: Oh that's a handy macro! Hacky but handy :laughing:

src/main.rs
mikeday: Yes that's one advantage of not needing to typecheck this yet :nerd_face: 

But I think with further experience we will probably refine these primitives into something more tractable. (An easier way to type it would just make it condense [[T]] into [T]).

PR #5, opened 2023-03-30
Extract evaluation and semantic domain into submodule

brendanzab:
Now that the _syntax_ has been decoupled from the _semantic domain_ (i.e. values) in #2, we can move all the evaluation-related code into a `semantics` submodule. This follows the general pattern I’ve been using to implement programming languages (see [brendanzab/language-garden](https://github.com/brendanzab/language-garden).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,39 +2,30 @@ use std::fs;
 use std::path::PathBuf;
 
 use clap::{Parser, ValueEnum};
-use serde::Serialize;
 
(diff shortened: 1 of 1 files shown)


mikeday:
Perhaps we could go halfway and make the Expr/Value split for now? :thinking:

brendanzab:
As in just have them all at the top level?

mikeday:
As a starting point yes, save a little indentation.

PR #8, opened 2023-03-30
Improve precision of matched bytes in lookahead

brendanzab:
The records all possible bytes matched in a hash set, letting us decode the full start-of-frame alternation. This could be made more efficient, but works for now.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,11 +1,70 @@
+use std::collections::HashSet;
 use std::env;
 use std::fs;
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Clone, Debug)]
 enum ByteSet {
-    Any,
-    Is(u8),
-    Not(u8),
+    Includes(HashSet<u8>),
+    Excludes(HashSet<u8>),
+}
+
+impl ByteSet {
+    pub fn any() -> ByteSet {
+        ByteSet::Excludes(HashSet::new())
+    }
+
+    pub fn is(b: u8) -> ByteSet {
+        ByteSet::Includes(HashSet::from([b]))
+    }
+
+    pub fn not(b: u8) -> ByteSet {
+        ByteSet::Excludes(HashSet::from([b]))
+    }
+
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -1,10 +1,66 @@
...
+                HashSet::is_disjoint(included0, included1)
+            }
+            // If the set of contained bytes are a subset of the excluded bytes,
+            // then the byte sets are disjoint
+            (ByteSet::Includes(included), ByteSet::Excludes(excluded))
+            | (ByteSet::Excludes(excluded), ByteSet::Includes(included)) => {
+                HashSet::is_subset(included, excluded)
+            }
+            // Too hard to enumerate all of these: just say that they aren’t disjoint
+            (ByteSet::Excludes(_), ByteSet::Excludes(_)) => false,
```
mikeday, line 43: Is there any reason not to check [0..255] ? Seems like it might be a one liner :thinking:

brendanzab: Not sure if I got it right!

src/main.rs
```diff
@@ -40,8 +40,11 @@ impl ByteSet {
             | (ByteSet::Excludes(excluded), ByteSet::Includes(included)) => {
                 HashSet::is_subset(included, excluded)
             }
-            // Too hard to enumerate all of these: just say that they aren’t disjoint
-            (ByteSet::Excludes(_), ByteSet::Excludes(_)) => false,
+            // Hard: enumerate these by brute force - they are disjoint if all
+            // bytes are contained exclusively in one byte set or the other set
```
mikeday, line 44: Does it need to be exclusive? I think they can be in both, they just have to be in at least one :thinking:

mikeday: Right, in order for two bytesets to be disjoint there must be one byte in both sets, which means there must be one byte that isn't excluded by _either_ set, but a byte excluded by _both_ sets is fine.

brendanzab:
Think this should be ready!

mikeday (approved):
Looks good to me!

PR #9, opened 2023-03-31
Use a non-allocating byte set

brendanzab:
This uses a specialised byte set as opposed to a `HashSet`. I'm wasn’t confident on some of the set operations, so added some property-based tests – but I can remove them we want to avoid adding a dependency at the moment.

### DIFF
--- a/src/byte_set.rs
+++ b/src/byte_set.rs
@@ -117,7 +117,20 @@ impl<const LEN: usize> From<[u8; LEN]> for ByteSet {
 
 impl fmt::Debug for ByteSet {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_set().entries(self.iter()).finish()
+        if self.len() < 128 {
+            f.debug_set().entries(self.iter()).finish()
+        } else {
+            f.write_str("!")?;
+            f.debug_set().entries((!self).iter()).finish()
+        }
+    }
+}
+
+impl ops::Not for &ByteSet {
+    type Output = ByteSet;
+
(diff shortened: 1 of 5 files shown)


brendanzab:
This should be ready to go. Realised over the weekend that I didn’t need the `Includes`/`Excludes` thing anymore, because that byteset covers the entire range of values!

mikeday:
Great, I'll take a look! Maybe we need a function for the common `Format::Byte(ByteSet::from(b))` idiom :sweat_smile:

src/byte_set.rs
```diff
@@ -0,0 +1,237 @@
+use std::{fmt, ops};
+
+/// Compact, allocation-free set of `u8`s.
+#[derive(Copy, Clone, PartialEq, Eq)]
```
mikeday, line 4: Should we add Debug traits everywhere? I've been adding them as needed because they sure are handy when debugging.

src/byte_set.rs
```diff
@@ -0,0 +1,237 @@
...
+pub struct ByteSet {
+    /// Each bit of the array represents a number in the range `0..=255`
+    bits: [u64; 4],
```
mikeday, line 7: So Rust has a u128 but not a u256, so close! :laughing:

src/byte_set.rs
brendanzab: I implemented it manually :)

mikeday: Oh so you did! Very neat :nerd_face:

mikeday (approved):
Looks great!

PR #10, opened 2023-04-02
Fix lookahead issues and add some tests

mikeday:


src/main.rs
```diff
@@ -283,11 +283,28 @@ impl Format {
...
+        match self {
+            Format::Fail => false,
+            Format::Empty => true,
+            Format::Byte(_) => false,
+            Format::Alt(a, b) => a.nullable() || b.nullable(),
+            Format::Cat(a, b) => a.nullable() && b.nullable(),
+            Format::Tuple(fields) => fields.iter().all(|f| f.nullable()),
+            Format::Record(fields) => fields.iter().all(|(_, f)| f.nullable()),
+            Format::Repeat(_a) => true,
+            Format::RepeatCount(_expr, _a) => true,
```
brendanzab, line 296: I guess we don’t count if the length might be non-zero? Makes sense for now! 👍

src/main.rs
```diff
@@ -464,50 +506,74 @@ impl Decoder {
...
                     dfields.push((name.clone(), df));
                 }
 
                 Ok(Decoder::Record(dfields))
             }
             Format::Repeat(a) => {
-                // FIXME next should be a|opt_next ?
+                if a.nullable() {
+                    return Err("cannot repeat nullable format".to_string());
+                }
```
brendanzab, line 560: This is nice!

src/main.rs
```diff
@@ -464,50 +506,74 @@ impl Decoder {
...
+                            f,
+                            Some(&Format::Cat(
+                                Box::new(Format::Tuple(remain.to_vec())),
+                                Box::new(next.clone()),
+                            )),
+                        )?,
                     };
-                    let df = Decoder::compile(f, opt_next)?;
                     dfields.push(df);
                 }
```
brendanzab, line 528: Might be a little cleaner to use [`std::slice::Iter::as_slice`](https://doc.rust-lang.org/std/slice/struct.Iter.html#method.as_slice):

```rust
let mut dfields = Vec::with_capacity(fields.len());
let mut fields = fields.iter();

while let Some(f) = fields.next() {
    let df = match (remain.is_empty(), opt_next) {
        (true, None) => Decoder::compile(f, None)?,
        (true, Some(next)) => Decoder::compile(f, Some(next))?,
        (false, None) => {
            Decoder::compile(f, Some(&Format::Tuple(fields.as_slice().to_vec())))?
        }
        (false, Some(next)) => Decoder::compile(
            f,
            Some(&Format::Cat(
                Box::new(Format::Tuple(fields.as_slice().to_vec())),
                Box::new(next.clone()),
            )),
        )?,
      };
      dfields.push(df);
}
```

src/main.rs
```diff
@@ -43,7 +43,7 @@ impl ByteSet {
             // Hard: enumerate these by brute force - they are disjoint if all
             // bytes are contained in one byte set or the other set
             (ByteSet::Excludes(excluded0), ByteSet::Excludes(excluded1)) => {
-                (0..=u8::MAX).all(|b| !excluded0.contains(&b) || !excluded1.contains(&b))
+                (0..=u8::MAX).all(|b| excluded0.contains(&b) || excluded1.contains(&b))
```
brendanzab, line 46: Oops!

brendanzab (approved):
Looks good!

src/main.rs
```diff
@@ -1047,3 +1113,359 @@ fn main() -> Result<(), Box<dyn std::error::Error + 'static>> {
...
+        rejects(&d, &[]);
+        rejects(&d, &[0x00]);
+    }
+
+    #[test]
+    fn compile_empty() {
+        let f = Format::Empty;
+        let d = Decoder::compile(&f, None).unwrap();
+        accepts(&d, &[], &[], Value::Unit);
+        accepts(&d, &[0x00], &[0x00], Value::Unit);
```
brendanzab, line 1159: I usually prefer to separate out independent test cases into separate test functions so they can fail independently, but it’s not a huge issue (happy to fix that myself). Good to have some tests for this.

src/main.rs
mikeday: Would the `remain.is_empty()` need to be a peek or is there some other way to check if it's the last one? Or just `fields.as_slice().is_empty()` :thinking:

brendanzab: Yeah I think that’s `.as_slice().is_empty()`. Maybe not as clean I dunno - and to be honest I think `Iter::as_slice` could have a clearer name.

brendanzab: Oh, you could just use `fields.is_empty()` (from `ExactSizeIterator`)

mikeday: Looks like that's still unstable, but len() == 0 works.

brendanzab (approved):

PR #11, opened 2023-04-03
DNL segment and subsequent scans can now be parsed

brendanzab:
As of #10 we can now compile these parts of JPEG


mikeday (approved):
Great!

PR #12, opened 2023-04-04
More work on JPEG and EXIF

brendanzab:
This adds a `WithRelativeOffset` format to parse more of EXIF. Not entirely sure of the name, and it’s a bit clunky to use. Adding and absolute offset format might make things easier.


mikeday (approved):
Looks good!

PR #13, opened 2023-04-05
Initial support for JSON output

brendanzab:
This lets us render decoded values as a JSON object. We might want to alter the schema later, just using the default [serde](https://serde.rs/) options at the moment. To actually work with the output values in a nicer way we might also want to output the format as a JSON object, along with the byte ranges for each node in the decoded values.


mikeday (approved):
Great!

PR #15, opened 2023-04-05
Render decoded values as a tree

brendanzab:
This prints values as a tree, which is a bit more readable. For example:

```
├── soi := 216
├── frame :=
│   ├── app0 :=
│   │   ├── marker := 224
│   │   ├── length := 16
│   │   └── data :=
│   │       ├── identifier :=
│   │       │   ├── 0 := 74
│   │       │   ├── 1 := 70
│   │       │   ├── 2 := 73
│   │       │   ├── 3 := 70
│   │       │   └── 4 := 0
│   │       ├── version-major := 1
│   │       ├── version-minor := 1
│   │       ├── density-units := 1
│   │       ├── density-x := 50
│   │       ├── density-y := 50
│   │       ├── thumbnail-width := 0
│   │       ├── thumbnail-height := 0
│   │       └── thumbnail-pixels := []
│   ├── segments :=
│   │   ├── 0 :=
│   │   │   ├── marker := 219
│   │   │   ├── length := 67
│   │   │   └── data :=
│   │   │       ├── precision-table-id := 0
│   │   │       └── elements :=
│   │   │           ├── 0 := 8
│   │   │           ├── 1 := 6
│   │   │           ├── 2 := 6
│   │   │           ├── 3 := 7
│   │   │           ├── 4 := 6
│   │   │           ├── 5 := 5
│   │   │           ├── 6 := 8
│   │   │           ├── 7 := 7
│   │   │           ├── 8 := 7
│   │   │           ├── 9 := 7
│   │   │           ~
│   │   │           └── 63 := 50
...
```

Long sequences of atomic values are elided to reduce bloat in the output. An interactive visualiser might be able to do a better job of this allowing you to expand these long sequences interactively.

In the future we might want to look into rendering the byte ranges of the nodes, and the format that each node was parsed with.


mikeday (approved):
Very cool!

PR #16, opened 2023-04-05
Switches!

mikeday:
(documentation to go here)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1417,6 +1417,33 @@ mod tests {
         assert!(Decoder::compile(&f, &Next::Empty).is_err());
     }
 
+    // This test takes over a minute to run
+    //#[test]
+    fn compile_alt_opt_ambiguous_slow() {
+        let alt = alts([
+            is_byte(0x00),
+            is_byte(0x01),
+            is_byte(0x02),
+            is_byte(0x03),
+            is_byte(0x04),
+            is_byte(0x05),
+            is_byte(0x06),
+            is_byte(0x07),
+        ]);
+        let rec = record([
+            ("0", alt.clone()),
+            ("1", alt.clone()),
+            ("2", alt.clone()),
+            ("3", alt.clone()),
+            ("4", alt.clone()),
+            ("5", alt.clone()),
+            ("6", alt.clone()),
+            ("7", alt.clone()),
+        ]);
+        let f = alts([rec.clone(), rec.clone()]);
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -237,315 +241,277 @@ impl Format {
...
         }
     }
-    /*
-        pub fn disjoint(a: &Self, b: &Self) -> bool {
-            for i in 0..std::cmp::min(a.pattern.len(), b.pattern.len()) {
-                if ByteSet::disjoint(&a.pattern[i], &b.pattern[i]) {
-                    return true;
-                }
+
+    fn add_next(&mut self, index: usize, depth: usize, next: &Next) -> Result<(), ()> {
```
brendanzab, line 274: Is the `depth` the number of bytes of lookahead?

src/main.rs
```diff
@@ -94,14 +96,15 @@ enum Format {
...
 }
 
-#[derive(Debug)]
-struct Lookahead {
-    pattern: Vec<ByteSet>,
+#[derive(Clone, Debug)]
+struct Switch {
+    accept: Option<usize>,
+    branches: Vec<(ByteSet, Switch)>,
 }
```
brendanzab, line 103: Is this sort of like a tree of lookahead? Keeping track of the index of variant matched in the `accept` field?

src/main.rs
```diff
@@ -237,315 +241,277 @@ impl Format {
...
+
+    fn build(branches: &[&Format], next: &Next) -> Option<Switch> {
+        const MAX_DEPTH: usize = 16;
+        for depth in 0..MAX_DEPTH {
+            if let Some(switch) = Switch::build0(depth, branches, next) {
+                return Some(switch);
+            }
+        }
+        None
+    }
```
brendanzab, line 402: I imagine this is trying `0..MAX_DEPTH` bytes of lookahead, trying a longer lookahead on failure?

src/main.rs
mikeday: Yes, each level of the tree switches on the next byte.

src/main.rs
mikeday: Yep! If none of the (disjoint) bytesets match or if there is no input remaining then the accept field indicates which if any of the switch cases succeeds.

src/main.rs
mikeday: Yes, we want to construct the shallowest possible tree as deeper trees can take a very long time to construct on some pathological input. There may still be inputs that exhibit awful time complexity, unfortunately.

PR #18, opened 2023-04-11
Use a single tuple expression instead of pairs and units

brendanzab:
Not sure if this is helpful at all! Just thought it might simplify some things.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,14 +10,17 @@ mod byte_set;
 
 #[derive(Clone, PartialEq, Debug, Serialize)]
 pub enum Value {
-    Unit,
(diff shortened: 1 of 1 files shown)


mikeday:
I'm thinking we might actually want to distinguish tuple values and sequence values :thinking:

brendanzab:
Yeah, this does that!

mikeday:
Oh sorry, I misread! Let me look a little more carefully :nerd_face:

mikeday (approved):
That looks good to me!

brendanzab:
Feel free to wait till merging the #16 as I assume there’ll be some conflicts.

PR #20, opened 2023-04-12
Support Repeat1 formats for one-or-more

mikeday:


src/main.rs
```diff
@@ -1417,8 +1441,8 @@ mod tests {
         assert!(Decoder::compile(&f, &Next::Empty).is_err());
     }
 
-    // This test takes over a minute to run
-    //#[test]
+    #[test]
+    #[ignore] // This test takes over a minute to run
```
adrianwong, line 1445: Very minor (feel free to ignore):

```suggestion
    #[ignore = "this test takes over a minute to run"]
```

brendanzab: Oooh I didn't know about that!

brendanzab (approved):
Looks good (I think Adrian’s suggestion is good too)

src/main.rs
mikeday: Ooh nice metadata!

brendanzab (approved):

PR #22, opened 2023-04-13
Add an end-of-input format

brendanzab:


src/main.rs
```diff
@@ -234,6 +237,7 @@ impl Format {
         match self {
             Format::Fail => false,
             Format::Empty => true,
+            Format::EndOfInput => false,
```
brendanzab, line 240: Unsure about this!

src/main.rs
```diff
@@ -303,6 +307,7 @@ impl Switch {
         match f {
             Format::Fail => Ok(()),
             Format::Empty => self.add_next(index, depth, next),
+            Format::EndOfInput => self.accept(index),
```
brendanzab, line 310: Does this make sense?

src/main.rs
```diff
@@ -71,6 +71,8 @@ enum Format {
     Fail,
     /// A format that matches the empty byte string
     Empty,
+    /// Matches if the end of the input has be reached
```
mikeday, line 74: s/be/been/

src/main.rs
mikeday: I think `EndOfInput` is nullable? Given that it's equivalent to `Empty` with the added constraint that it must occur at the end of the input. So for example `Repeat(EndOfInput)` at the end of the file would present a problem!

mikeday: Thinking about it some more yes it's definitely nullable as `EndOfInput` can match the empty string.

src/main.rs
mikeday: Yes I think this makes sense.

mikeday (approved):
Looks good! Would be nice to have a test for `Cat(Repeat(0x00), EndOfInput)`

src/main.rs
brendanzab: I added a doc comment that hopefully clarifies the purpose of this method some more

brendanzab:
Test added, hopefully the method rename is ok too!

src/main.rs
```diff
@@ -1532,6 +1579,20 @@ mod tests {
         );
     }
 
+    #[test]
+    fn compile_cat_end_of_input() {
+        let f = Format::Cat(Box::new(is_byte(0x00)), Box::new(Format::EndOfInput));
```
mikeday, line 1584: Oh I was thinking this could use Repeat to check that compilation works and the loop knows when to stop

mikeday:
Yep the rename is fine!

PR #23, opened 2023-04-13
Clean up tree rendering with a context struct

brendanzab:
This was useful as I was starting to look into format-directed printing


mikeday (approved):
Looks good.

PR #24, opened 2023-04-13
Separate values from exprs

brendanzab:
This is is a half-way step towards #5.

Closes #5.


mikeday (approved):
Looks good! Simplifies things. :nerd_face:

PR #26, opened 2023-04-14
Labelled union formats

brendanzab:
This alters union and match formats so that they “remember” the branch that was matched in a labelled variant type.


brendanzab:
Just posting to show some stuff I was experimenting with. This is gonna be pretty slow as we’re now using strings a ton more (constructing and comparing them), especially in the JPEG scan data.

PR #28, opened 2023-04-16
Rename switch formats and add switch expressions

mikeday:


src/main.rs
```diff
@@ -21,6 +21,10 @@ pub enum Value {
 
 impl Value {
     const UNIT: Value = Value::Tuple(Vec::new());
+
+    fn from_bytes(bs: &[u8]) -> Value {
+        Value::Seq(bs.iter().copied().map(|b| Value::U8(b)).collect())
```
adrianwong, line 26: Very minor:

```suggestion
        Value::Seq(bs.iter().copied().map(Value::U8).collect())
```

PR #29, opened 2023-04-17
Add binary STL format

brendanzab:
Note that binary STL can’t really be disambiguated from the other formats automatically, so an explicit `--format=stl` argument needs to be provided.


mikeday:
I'm tempted to only support formats that can be identified without user intervention, is there nothing at all standard in the STL file header?

brendanzab:
Sadly no… I can shelve this one for now?

mikeday:
Might be best, maybe we can support FBX one day :sweat_smile:

PR #30, opened 2023-04-17
Use match formats to implement application segments

brendanzab:
I was a bit confused by the switch formats in #28, so had a go at implementing something similar using pattern matching instead. This also links in with the changes I was making in #26 to add variant types.

Closes #28.


src/main.rs
```diff
@@ -130,8 +177,8 @@ enum Format {
     WithRelativeOffset(Expr, Box<Format>),
     /// Transform a decoded value with a function
     Map(Func, Box<Format>),
-    /// Conditional format
-    If(Expr, Box<Format>, Box<Format>),
+    /// Pattern match on an expression
+    Match(Expr, Vec<(Pattern, Format)>),
```
brendanzab, line 181: Documentation could be better here… it was hard to document this (mentioning the the byte strings matched by the format), because this format is matching on an expression as well 😅

mikeday:
Interesting! The pattern matching looks great, I'll just have a little think about the implications for the Decoder.

brendanzab:
Yeah np, wasn‘t too hard to implement so thought I’d give it a go. Not sure if it runs into any issues with what you were planning for that.

src/main.rs
```diff
@@ -130,8 +177,8 @@ enum Format {
     WithRelativeOffset(Expr, Box<Format>),
     /// Transform a decoded value with a function
     Map(Func, Box<Format>),
-    /// Conditional format
-    If(Expr, Box<Format>, Box<Format>),
```
mikeday, line 134: Although Match does subsume If but I think If might still be more succinct for matching on booleans! :sweat_smile:

src/main.rs
```diff
@@ -140,19 +187,14 @@ struct MatchTree {
...
 /// Decoders with a fixed amount of lookahead
 enum Decoder {
     Fail,
     Empty,
     EndOfInput,
     Byte(ByteSet),
-    If(Cond, Box<Decoder>, Box<Decoder>),
-    Switch(MatchTree, Vec<Decoder>),
+    Alt(MatchTree, Box<Decoder>, Box<Decoder>),
+    Union(MatchTree, Vec<Decoder>),
```
mikeday, line 197: Maybe we can drop Alt altogether and rename Decoder::Union to Branch, to emphasise it's deterministic?

src/main.rs
brendanzab: Used a helper function instead!

src/main.rs
```diff
@@ -477,10 +504,10 @@ impl Decoder {
...
             Format::Byte(bs) => Ok(Decoder::Byte(*bs)),
             Format::Alt(a, b) => {
-                let da = Box::new(Decoder::compile(a, next)?);
-                let db = Box::new(Decoder::compile(b, next)?);
-                if let Some(tree) = MatchTree::build(&[a, b], next) {
-                    Ok(Decoder::If(Cond::MatchTree(tree), da, db))
+                let da = Decoder::compile(a, next)?;
+                let db = Decoder::compile(b, next)?;
+                if let Some(switch) = MatchTree::build(&[a, b], next) {
+                    Ok(Decoder::Branch(switch, vec![da, db]))
```
mikeday, line 510: Use `tree` instead of `switch` here I think.

mikeday (approved):
Looks good to me!

src/main.rs
brendanzab: Oops, missed this!

brendanzab: Pushed a fix to `main`

PR #32, opened 2023-04-17
Build match trees breadth first and cache paths for speed

mikeday:
This fixes all (?) the slow test cases and allows us to increase MAX_DEPTH.


brendanzab:
Does this mean we can now remove `#[ignore]` from the slow tests?

PR #33, opened 2023-04-17
Rename some switch types

mikeday:
Factored out these changes for convenience.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -135,14 +135,14 @@ enum Format {
 }
 
 #[derive(Clone, Debug)]
-struct Switch {
+struct MatchTree {
     accept: Option<usize>,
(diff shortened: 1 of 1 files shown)


brendanzab (approved):

src/main.rs
```diff
@@ -135,14 +135,14 @@ enum Format {
 }
 
 #[derive(Clone, Debug)]
-struct Switch {
+struct MatchTree {
```
brendanzab, line 138: Maybe some other options could be `LookaheadTree` or `ByteTree`?

mikeday: Yes, I think we're going to need other lookahead mechanisms sooner or later and might adjust names further then.

PR #34, opened 2023-04-18
Format-directed printing

brendanzab:
This uses formats to guide the printing of decoded values when using `--output=tree`. For some map formats with tricky functions (like `Func::Match` and `Match::Stream`) we bail and fall back to non-format-directed printing. This can be seen in the JPEG scan data, for example.


mikeday (approved):
Looks good!

PR #35, opened 2023-04-18
Remove some redundant formats

brendanzab:
Not sure if you’d rather keep these, but I noticed when implementing the format-directed printing that we have some redundant formats that could be removed.


mikeday (approved):
Yes that looks reasonable.

PR #39, opened 2023-04-25
Special case the display of some ASCII strings

mikeday:


frontend/index.html
```diff
@@ -65,11 +65,15 @@
 
     li:last-child,
     dd dt:last-of-type,
-    li > * > :first-child {
+    li > :not(span) > :first-child {
```
sashaboyd, line 68: Can we use :not(.text) here instead? It's a bit more descriptive.

sashaboyd (approved):

PR #40, opened 2023-04-26
Begin GIF format

brendanzab:
Begins implementing the GIF format - just up to the global colour table for now. I also sorted the formats alphabetically to keep things clean.


mikeday (approved):
Looks good!

PR #42, opened 2023-04-26
Add some more operators

brendanzab:
This removes the need for a special `is-even` operation, and lets us use an `if` format for GIF’s global color table.


mikeday (approved):
Looks good!

PR #43, opened 2023-04-26
Fix clippy lints

brendanzab:
This fixes all clippy lint warnings. To reduce annoyance I selectively disabled the redundant clone warnings when constructing the formats, and in the tests.


mikeday (approved):
Looks good!

PR #46, opened 2023-04-28
Follow GIF89a grammar more closely

brendanzab:
This more closely enforces the order that blocks appear in the GIF format, as per Appendix B of the [GIF89a specification](https://www.w3.org/Graphics/GIF/spec-gif89a.txt). [What's In A GIF - Bit by Byte](https://www.matthewflickinger.com/lab/whatsinagif/bits_and_bytes.asp) also provides a good overview of this.


mikeday (approved):
Looks good.

PR #48, opened 2023-05-01
Attempt to improve the input manipulation formats

brendanzab:
This extracts the notion of “not consuming the input” from the input manipulation formats, and adjusts the names of these formats to work more nicely together.

This was partly inspired by the paper [“Interval Parsing Grammars for File Format Parsing”](https://arxiv.org/abs/2304.04859) - except here we default to consuming bytes from the input as opposed to defaulting to parsing in-place.


src/lib.rs
```diff
@@ -210,10 +210,14 @@ pub enum Format {
     Repeat1(Box<Format>),
     /// Repeat a format an exact number of times
     RepeatCount(Expr, Box<Format>),
-    /// Restrict a format to a sub-stream of a given number of bytes
-    Slice(Expr, Box<Format>),
-    /// Matches a format at a byte offset relative to the current stream position
-    WithRelativeOffset(Expr, Box<Format>),
+    /// Matches a format without consuming the input
+    WithInput(Box<Format>),
```
brendanzab, line 214: Not quite happy with the name of this format.

mikeday: Interesting, so instead of say `Intersection(A, B)` there would be `Cat(WithInput(A), B)` :thinking:

src/lib.rs
```diff
@@ -210,10 +210,14 @@ pub enum Format {
...
+    /// Takes a given number of bytes from the start of the input, matching
+    /// the format against those bytes
+    TakeBytes(Expr, Box<Format>),
+    /// Drops a given number of bytes from the start of the input, matching
+    /// the format against the remaining input
+    DropBytes(Expr, Box<Format>),
```
mikeday, line 220: `DropBytes` seems unnecessarily parallel to `TakeBytes`, seems like it doesn't need the Format argument and could just skip N bytes? Whereas `TakeBytes` does require the Format argument.

mikeday:
Not sure about the names here! But extracting `WithInput` from `WithRelativeOffset` is thought provoking :thinking:

src/lib.rs
brendanzab: Oh right, `TakeBytes` can be used to implement something like `DropBytes`?
```text
def drop-bytes (n : Size) (f : Format) :=
  map _.data {
    skipped <- take-bytes n {},
    data <- f,
  };
```
Hadn’t thought of that! 🤔

src/lib.rs
brendanzab: Yeah, maybe intersection would be better… not sure!

brendanzab:
Yeah not sold on this, thought I’d post in case it was thought provoking! :)

PR #49, opened 2023-05-02
Implement top-level items

brendanzab:
This implements top-level items, which makes the tree-output more readable, and moves us closer to being able to define a surface syntax for the data description language. We still can’t express parameterised formats, which means that there are some remaining formats that don’t show up in the decoded output.

Because we cannot know the `next` format ahead of time, all top-level items are inlined in the compiled decoders. This is not a massive issue right now, but could result in code bloat, and could impact future code generation.


mikeday:
Exciting! Does this affect the HTML frontend at all? Can we expose the top-level item names in the JSON?

src/lib.rs
```diff
@@ -224,6 +226,35 @@ impl Format {
...
+            names: Vec::new(),
+            formats: Vec::new(),
+        }
+    }
+
+    pub fn define_format(&mut self, name: impl Into<String>, format: Format) -> Format {
+        let level = self.names.len();
+        self.names.push(name.into());
+        self.formats.push(format);
+        Format::ItemVar(level)
```
mikeday, line 246: Interesting that this is called "level", suggests nesting? :thinking:

src/main.rs
```diff
@@ -113,75 +120,84 @@ fn not_byte(b: u8) -> Format {
...
-fn u16be() -> Format {
-    Format::Map(
-        Func::U16Be,
-        Box::new(Format::Tuple(vec![any_byte(), any_byte()])),
-    )
-}
+    let u16be = module.define_format(
+        "base.u16be",
+        Format::Map(Func::U16Be, Box::new(tuple([u8.clone(), u8.clone()]))),
+    );
```
mikeday, line 152: Wow we have u16be in a prelude now that's neat :joy:

mikeday (approved):
Looks very cool!

mikeday:
I think the inlining in the decoder is okay, factoring that out is actually a little bit challenging given that compiling a format can be affected by what follows it.

src/lib.rs
brendanzab: Level in the sense of a de Bruijn level as opposed to a de Bruijn index. This is in contrast to `Expr::Var` which uses a de Bruijn index.

brendanzab:
> Exciting! Does this affect the HTML frontend at all? Can we expose the top-level item names in the JSON?

This doesn’t affect the HTML frontend yet - to take advantage of this we’d need the frontend to use the format to direct rendering, like in `doodle::output::tree::print_decoded_value`.

mikeday:
Oh right, maybe it's time to provide the combination of format + value to the HTML interface.

PR #51, opened 2023-05-05
Move subformats into separate files

brendanzab:
We’re going to be adding more formats so it probably makes sense to break things up a bit to keep things manageable.


mikeday (approved):
Looks good!

PR #53, opened 2023-05-07
Some more Expr ops (Add, BitOr, TupleProj)

mikeday:


src/output/tree.rs
```diff
@@ -322,6 +332,10 @@ impl<'module, W: io::Write> Context<'module, W> {
 
     fn write_proj_expr(&mut self, expr: &Expr) -> io::Result<()> {
         match expr {
+            Expr::TupleProj(head, index) => {
+                self.write_proj_expr(head)?;
+                write!(&mut self.writer, "[{}]", index)
+            }
```
brendanzab, line 338: Maybe (for consistency with how we render `Func::TupleProj`):

```suggestion
            Expr::TupleProj(head, index) => {
                self.write_proj_expr(head)?;
                write!(&mut self.writer, ".{index}")
            }
```

brendanzab (approved):
Looks good!

src/output/tree.rs
mikeday: Done! I always forget that variables can be used in fmt macros like that, it's neat.

PR #54, opened 2023-05-08
Format::MapExpr

mikeday:
Expr is more powerful than Func and in fact can even replace `Func::U16Be` and more. This `MapExpr` format could almost replace `Map` entirely (aside from `Func::Stream`), however it currently does not print out names correctly as it pushes another item on the stack that the output tree does not know about. How to fix this?

### DIFF
--- a/src/output/tree.rs
+++ b/src/output/tree.rs
@@ -97,6 +97,7 @@ impl<'module, W: io::Write> Context<'module, W> {
             Format::Map(Func::U32Be, _) => self.write_value(value),
             Format::Map(Func::U32Le, _) => self.write_value(value),
             Format::Map(Func::Stream, _) => self.write_value(value),
+            Format::MapExpr(_expr, _) => self.write_value(value),
             Format::Match(head, branches) => {
                 let head = head.eval(&self.values);
                 let initial_len = self.values.len();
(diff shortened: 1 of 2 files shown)


src/output/tree.rs
```diff
@@ -422,6 +423,14 @@ impl<'module, W: io::Write> Context<'module, W> {
...
             }
 
+            Format::MapExpr(expr, format) => {
+                write!(&mut self.writer, "map (")?;
+                self.write_expr(expr)?;
+                write!(&mut self.writer, ")")?;
+                write!(&mut self.writer, " ")?;
+                self.write_atomic_format(format)
+            }
+
```
brendanzab, line 433: We’ll want to push a name onto the context, as we do in `Context::write_record`. We should also push a value as well, but we don’t currently have a way of representing open values yet. It shouldn’t matter in the short term, but it would probably be a good idea to add a FIXME about that.

```suggestion
            Format::MapExpr(expr, format) => {
                let name = "x";
                write!(&mut self.writer, "map (fun {name} => ")?;
                self.names.push(name.to_string());
                // FIXME: push value
                self.write_expr(expr)?;
                self.names.pop();
                write!(&mut self.writer, ") ")?;
                self.write_atomic_format(format)
            }

```

brendanzab:
One issue with a general expression like this is that it’s harder to deal with when doing format directed printing (as in `write_decoded_value`). I’m very curious if there are clever approaches we could use to cope with this.

brendanzab:
Another option for moving in the direction of this might be to bind a variable when evaluating `Func::Expr`.

src/output/tree.rs
mikeday: Thanks, I've made that change and it makes sense!

mikeday:
I think this MapExpr format is useful and usable, so probably worth merging even though it does raise issues for display of the output that remain to be addressed.

brendanzab (approved):

PR #56, opened 2023-05-11
Combine the functionality of map formats

brendanzab:
Building off #54, this moves the `Func` operations into `Expr`, meaning we no longer need two map formats.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -121,6 +121,7 @@ pub enum Expr {
     RecordProj(Box<Expr>, String),
     Variant(String, Box<Expr>),
     Seq(Vec<Expr>),
+    Match(Box<Expr>, Vec<(Pattern, Expr)>),
 
     BitAnd(Box<Expr>, Box<Expr>),
     BitOr(Box<Expr>, Box<Expr>),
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -130,6 +131,12 @@ pub enum Expr {
     Shl(Box<Expr>, Box<Expr>),
     Add(Box<Expr>, Box<Expr>),
     Sub(Box<Expr>, Box<Expr>),
+
+    FromU16Be(Box<Expr>),
+    FromU16Le(Box<Expr>),
+    FromU32Be(Box<Expr>),
+    FromU32Le(Box<Expr>),
```
mikeday, line 138: The names of these are a bit funky aren't they, because `FromU32Be` means "applies to a tuple of four U8 values and returns a single U32 value by interpreting them in big endian order", it feels like the "FromU32" part of the name is almost a misdirection given that it's also "From4U8" and "ToU32"!

mikeday (approved):
Looks good!

src/lib.rs
brendanzab: What about `U32FromBe`?

brendanzab: I can do that for now! Though I feel like eventually it would be nice to have a name distinct from the format name.

PR #58, opened 2023-05-17
Format::RepeatUntil.

mikeday:
Intended for reading sequences of items until a condition is satisfied by the last item, for example a sequence of blocks where the last block has "final = true".

### DIFF
--- a/src/output/tree.rs
+++ b/src/output/tree.rs
@@ -70,12 +70,13 @@ impl<'module, W: io::Write> Context<'module, W> {
                 Value::Record(value_fields) => self.write_record(value_fields, Some(format_fields)),
                 _ => panic!("expected record"),
             },
-            Format::Repeat(format) | Format::Repeat1(format) | Format::RepeatCount(_, format) => {
-                match value {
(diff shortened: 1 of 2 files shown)


src/output/tree.rs
```diff
@@ -70,12 +70,13 @@ impl<'module, W: io::Write> Context<'module, W> {
...
-                    _ => panic!("expected sequence"),
-                }
-            }
+            Format::Repeat(format)
+            | Format::Repeat1(format)
+            | Format::RepeatCount(_, format)
+            | Format::RepeatUntil(_, format) => match value {
+                Value::Seq(values) => self.write_seq(values, Some(format)),
+                _ => panic!("expected sequence"),
+            },
```
brendanzab, line 79: Also need to add a case in `write_format`

mikeday: Oops! Always forget that one, gave me an infinite loop last time :sweat_smile:

brendanzab (approved):

brendanzab:
Looks good!

PR #59, opened 2023-05-17
Tuple should push its fields on the stack while parsing, just like Record

mikeday:


brendanzab:
Ahh, I changed this because I was assuming that tuples would be non-dependent. What would the surface syntax look like for this kind of format? I guess each element of the tuple would bind a name?

brendanzab:
We’d need to change the tree formatter as well to push the values, as in the record formats. This will necessitate splitting the tuple rendering from the sequence rendering.

mikeday:
Hmm didn't think about the naming, good point! I might not have thought this one through, it just bit me when I was trying to make a dependent tuple, but it probably should have been a record yeah.

PR #60, opened 2023-05-18
gzip/deflate formats

mikeday:
A work in progress that can parse gzip files using fixed or dynamic huffman encoding.


brendanzab:
Could perhaps add some new tests to `tests/decode.rs` for the added sample files?

PR #61, opened 2023-05-18
Do a better job of recovering mapped formats

brendanzab:
This handles the possibility of multiple projections in map formats

### DIFF
--- a/src/output/tree.rs
+++ b/src/output/tree.rs
@@ -21,6 +21,28 @@ fn is_atomic_value(value: &Value) -> bool {
     }
 }
 
+/// Attempt to recover a format that can be used to render a value that was
+/// decoded with a map format. We currently handle cases like:
+///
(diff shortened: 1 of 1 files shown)


mikeday:
I'm curious what effect this has on the output, currently mapped formats confuse me a little

mikeday:
Oops hit close on the pull request :facepalm:

brendanzab:
It doesn’t have any effect on the current mapped output as we don’t have any cases of formats like `map (fun x => x.foo.bar.3) format` yet. Currently we only handle `map (fun x => x.foo) format` and `map (fun x => x.1) format`.

brendanzab:
Updated the `mapped_format` function with a doc comment that hopefully clarifies things more

mikeday (approved):
Looks good!

PR #76, opened 2023-10-20
Add Expr::Div and Expr::Shr

archaephyrryx:
Adds `Expr` primitives for the Divide and Right-Shift operations, with support for pretty-printing said operations.

Clarifies documentation of `Format::Slice` to indicate its actual behavior under the current implementation (i.e. skip any leftover bytes within the slice after consuming as many bytes as the wrapped Format can allow it to).


mikeday (approved):
That looks good!

PR #77, opened 2023-10-23
Tar format

archaephyrryx:
Adds tar format, along with new variants, methods, and miscellaneous changes to support Tar-specific needs.


mikeday (approved):
Looks good!

PR #82, opened 2023-10-26
Monoidal printing model with precedence rules

archaephyrryx:
Introduces a parallel model for Pretty-Printing in the Tree-style, using a Writer-free `MonoidalPrinter` that builds text `Fragments` using structural induction. This model also uses precedence-based rules for parenthesizing expressions and formats.

WIP because this has not been tested rigorously enough, and because there is not yet a clean baton-pass from the previous `Context` model to `MonoidalPrinter`.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -44,6 +44,11 @@ impl Fragment {
         matches!(self, &Fragment::Empty)
     }
 
+    /// Forms a compound fragment from a Fragment-valued iterable, with
+    /// an optional Fragment separating each element in the output sequence.
+    ///
+    /// It is more efficient to pass `sep := None` than `sep := Some(Fragment::Empty)`,
+    /// but the resulting output will differ in performance alone, and not output.
     fn seq(items: impl IntoIterator<Item = Fragment>, sep: Option<Fragment>) -> Self {
         Self::Sequence {
             items: items.into_iter().collect(),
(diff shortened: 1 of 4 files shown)


src/output.rs
```diff
@@ -1,2 +1,231 @@
...
+}
+
+#[derive(Clone, Default)]
+pub enum Fragment {
+    #[default]
+    Empty,
+    Symbol(Symbol),
+    Char(char),
+    String(std::borrow::Cow<'static, str>),
+    DisplayAtom(std::rc::Rc<dyn std::fmt::Display>),
```
mikeday, line 33: I think can import Cow and Rc directly rather than qualifying them everywhere.

archaephyrryx: Done.

src/output/tree.rs
```diff
@@ -816,3 +828,1358 @@ impl<'module, W: io::Write> Context<'module, W> {
...
+    const ATOM: Self = Precedence::Atomic;
+
+    const FORMAT_COMPOUND: Self = Self::Top;
+    const FORMAT_ATOM: Self = Self::Atomic;
+
+    fn bump_format(&self) -> Self {
+        match self {
+            Precedence::Top => Precedence::Atomic,
+            Precedence::Atomic => Precedence::Atomic,
+            _ => unreachable!("Unexpected non-format precdence level {self:?}"),
```
mikeday, line 1935: "precedence"

mikeday (approved):
Looks fine to me!

src/output/tree.rs
archaephyrryx: Done.

PR #84, opened 2023-10-30
Add Base formats for ascii digits (octal, hex, decimal)

archaephyrryx:


src/bin/doodle/format/base.rs
```diff
@@ -1,5 +1,5 @@
 use doodle::byte_set::ByteSet;
-use doodle::{Expr, Format, FormatModule, FormatRef, Pattern};
+use doodle::{ Expr, Format, FormatModule, FormatRef, Pattern };
```
mikeday, line 2: More unexpected formatting changes here 🤔

archaephyrryx: Indeed. Not sure if reverting this change and re-running `rustfmt` makes a difference, however.

mikeday (approved):
Looks good to me!

PR #87, opened 2023-11-02
Fixes for tar, pretty-printing

archaephyrryx:
Fixes various minor nits in the tar format definition, adjusts pretty-printing rules to handle more combinatorial Formats as ASCII string/characters, and similarly to treat more combinatorial Formats as "implied values" (i.e. can be omitted from tree printing)

### DIFF
--- a/src/output/tree.rs
+++ b/src/output/tree.rs
@@ -806,21 +806,25 @@ impl<'module, W: io::Write> Context<'module, W> {
             Format::EndOfInput => write!(&mut self.writer, "end-of-input"),
             Format::Align(n) => write!(&mut self.writer, "align {n}"),
 
-            Format::Byte(bs) => {
-                if bs.len() < 128 {
+            Format::Byte(bs) => match bs.len() {
+                0 => unreachable!("matches against the empty byteset are unsatisfiable"),
+                1..=127 => {
                     write!(&mut self.writer, "[=")?;
                     for b in bs.iter() {
                         write!(&mut self.writer, " {b}")?;
                     }
                     write!(&mut self.writer, "]")
-                } else {
+                }
+                128..=255 => {
                     write!(&mut self.writer, "[!=")?;
                     for b in (!bs).iter() {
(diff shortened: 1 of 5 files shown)


src/output/tree.rs
```diff
@@ -840,6 +837,11 @@ impl<'module, W: io::Write> Context<'module, W> {
     }
 }
 
+#[inline]
+fn name_is_ascii_string(name: &str) -> bool {
+    name.contains("ascii") && name.contains("string")
+}
+
```
mikeday, line 848: Not sure about this rule :thinking:

archaephyrryx: It is the cleanest way I saw to get bespoke ascii-strings working as ascii string. Consider `tar.rs` `tar.ascii-string(.opt0(.nonempty)?)?` How should we handle those without adding a new override for every format?

archaephyrryx: Alternatively, we could have an `AsAsciiString` format wrapper that makes whatever s inside of it treated as though ascii, but that could get us in trouble due to weird constructions.

archaephyrryx: It could be narrowed to `name.contains("asciiz-string") || name.contains("ascii-string")` if you like...

archaephyrryx: Which would satisfy all the current passing cases while filtering out a whole class of potential false positives down the line.

mikeday: Perhaps using `ascii-char` would be more flexible, but perhaps we would need `ascii-char-non-nul`?

archaephyrryx: Probably so.

archaephyrryx: Possibly also an explicit `ascii-char.non-null` and `ascii-char.null` to keep the same prefix match for the ItemVar names.

archaephyrryx: Relegated to issue #88

mikeday (approved):
Looks good to me!

PR #90, opened 2023-11-02
Add Format::PeekNot and improve switch building for Slice, RepeatCount, and Peek.

mikeday:


src/lib.rs
```diff
@@ -1278,203 +1290,360 @@ impl Format {
...
+        }
+    }
+
+    pub fn add(module: &'a FormatModule, f: &'a Format, next: Rc<Next<'a>>) -> MatchTreeStep<'a> {
+        match f {
+            Format::ItemVar(level, _args) => Self::add(module, module.get_format(*level), next),
+            Format::Fail => Self::reject(),
+            Format::EndOfInput => Self::accept(),
+            Format::Align(_) => {
+                Self::accept() // FIXME
```
archaephyrryx, line 1485: How difficult would it be to implement Align? Is it possible under the current model but just tedious, or is it actually not possible as designed?

archaephyrryx (approved):
Looks good to me!

src/lib.rs
mikeday: Align could be implemented by treating it as equivalent to some optional number of arbitrary bytes, but there's little value in doing that right now as we only use it in one place in the deflate format and it's not something likely to be switched upon.

It's similar to the slightly more likely case of the upper bound for RepeatCount, which is also ignored. For example if you have RepeatCount(expr, f) where the bounds for expr are min..max then currently we treat that as being min repetitions of f, but if max > min then we just stop at that point instead of exploring every possible value that max could take, as this isn't necessary for any format. Align(n) could be implemented as `RepeatCount(expr, byte)` where expr has bounds from 0..n-1 depending on the current offset.

Anyway I think we can ignore Align for now!

PR #91, opened 2023-11-02
Drop tree::Context and comparison tests

archaephyrryx:
Drops `tree.rs` `Context` pretty-printer model, as well as the tests that were comparing its output against `MonoidalPrinter`.


mikeday (approved):
That looks fine.

PR #95, opened 2023-11-03
Tar format patches, improvements to MonoidalPrinter

archaephyrryx:
Fixes bug in Tar format specification that anticipates an errant trailing all-empty block after file contents if filesize is divisible by 512. This change has the side-effect of collating file bytes within the tar format into a single byte sequence, rather than a sequence of 512-byte blocks.

Additionally adds a new `Flag` value `hide_double_underscore_fields` that, when enabled (true by default), elides the display of any record-field whose name matches `/^__.*/`. This enables format-specific elision of certain uninteresting fields, rather than just merely eliding their values if implicit.


src/bin/doodle/format/tar.rs
```diff
@@ -37,14 +37,27 @@ pub fn main(module: &mut FormatModule, base: &BaseModule) -> FormatRef {
...
 
-    // A format for uninterpreted N-byte values
-    let cbytes = |len: u16| -> Format { repeat_count(Expr::U16(len), base.u8()) };
+    let nul_or_wsp = module.define_format(
+        "tar.padding-char",
+        Format::Byte(ByteSet::from([0x00, b' '])),
+    );
 
-    const MAGIC: &[u8; 6] = b"ustar\x00";
+    // A format for uninterpreted N-byte values
```
mikeday, line 47: Are these uninterpreted N-byte values? It appears to be an octal string :thinking:

mikeday (approved):

src/bin/doodle/format/tar.rs
archaephyrryx: The comment is perhaps a bit out-of-date. I will do a bit of digging to re-confirm that only octal digits are allowed, rename the binding, and amend the comment.

archaephyrryx: I have attempted to solve the issue, let me know if it is better now.

PR #97, opened 2023-11-06
Assorted tweaks to `base.rs` and ByteSet

archaephyrryx:
- Adds `repeat_byte` format helper function to `base.rs` when a particular byte is repeated a fixed number of times.
- Adds `VALID_ASCII : ByteSet` const value for future use
- Fixes off-by-one in offset comments of tar header
- Adds `#[inline]` attributes, const modifiers to selected ByteSet functions (particularly, so that `from_bits`, `empty`, and `full` can be used in `const` declarations)
- Adds optimized `From<RangeInclusive<u8>>` for ByteSet, as well as `From<&' [u8]>` and `FromIterator<u8>`.

### DIFF
--- a/src/bin/doodle/format/base.rs
+++ b/src/bin/doodle/format/base.rs
@@ -78,7 +78,7 @@ pub fn is_bytes(bytes: &[u8]) -> Format {
 }
 
 /// ByteSet consisting of 0..=127, or the valid ASCII range (including control characters)
-pub const VALID_ASCII : ByteSet = ByteSet::from_bits([u64::MAX, u64::MAX, 0, 0]);
+pub const VALID_ASCII: ByteSet = ByteSet::from_bits([u64::MAX, u64::MAX, 0, 0]);
 
 pub struct BaseModule {
     bit: FormatRef,
--- a/src/bin/doodle/format/base.rs
+++ b/src/bin/doodle/format/base.rs
@@ -65,6 +65,10 @@ pub fn is_byte(b: u8) -> Format {
     Format::Byte(ByteSet::from([b]))
 }
 
+pub fn repeat_byte(count: u32, b: u8) -> Format {
+    Format::RepeatCount(Expr::U32(count), Box::new(is_byte(b)))
+}
+
 pub fn not_byte(b: u8) -> Format {
     Format::Byte(!ByteSet::from([b]))
 }
@@ -72,6 +76,9 @@ pub fn not_byte(b: u8) -> Format {
 pub fn is_bytes(bytes: &[u8]) -> Format {
     tuple(bytes.iter().copied().map(is_byte))
 }
+
+/// ByteSet consisting of 0..=127, or the valid ASCII range (including control characters)
+pub const VALID_ASCII : ByteSet = ByteSet::from_bits([u64::MAX, u64::MAX, 0, 0]);
 
 pub struct BaseModule {
     bit: FormatRef,
--- a/src/byte_set.rs
+++ b/src/byte_set.rs
@@ -139,6 +139,12 @@ impl<'a> From<&'a [u8]> for ByteSet {
     }
 }
 
+impl From<ops::Range<u8>> for ByteSet {
+    fn from(value: ops::Range<u8>) -> Self {
+        ByteSet::from(value.start..=value.end - 1)
+    }
+}
+
 impl From<ops::RangeInclusive<u8>> for ByteSet {
     fn from(value: ops::RangeInclusive<u8>) -> Self {
         // because the values are adjacent, we can optimize if they are within the same quadrant
--- a/src/byte_set.rs
+++ b/src/byte_set.rs
@@ -11,19 +11,22 @@ pub struct ByteSet {
 }
 
 impl ByteSet {
-    pub fn new() -> ByteSet {
+    #[inline(always)]
+    pub const fn new() -> ByteSet {
         ByteSet::empty()
     }
 
-    pub fn from_bits(bits: [u64; 4]) -> ByteSet {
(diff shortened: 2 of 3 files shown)


src/bin/doodle/format/base.rs
```diff
@@ -73,6 +77,9 @@ pub fn is_bytes(bytes: &[u8]) -> Format {
     tuple(bytes.iter().copied().map(is_byte))
 }
 
+/// ByteSet consisting of 0..=127, or the valid ASCII range (including control characters)
+pub const VALID_ASCII: ByteSet = ByteSet::from_bits([u64::MAX, u64::MAX, 0, 0]);
```
mikeday, line 81: Temping to say "7-bit ASCII" somehow rather than "valid", specifically... :thinking:

src/byte_set.rs
```diff
@@ -11,19 +11,22 @@ pub struct ByteSet {
 }
 
 impl ByteSet {
-    pub fn new() -> ByteSet {
+    #[inline(always)]
```
mikeday, line 14: Is this inline declaration necessary for use in const? Is the const keyword not enough?

src/byte_set.rs
```diff
@@ -126,21 +133,55 @@ impl<const LEN: usize> From<[u8; LEN]> for ByteSet {
...
 }
 
-impl From<RangeInclusive<u8>> for ByteSet {
-    fn from(range: RangeInclusive<u8>) -> ByteSet {
-        let mut bs = ByteSet::new();
-        for b in range {
-            bs.insert(b);
+impl From<ops::Range<u8>> for ByteSet {
+    fn from(value: ops::Range<u8>) -> Self {
+        ByteSet::from(value.start..=value.end - 1)
```
mikeday, line 148: Will this behave on empty ranges? In particular 0..0 :thinking:

src/byte_set.rs
```diff
@@ -126,21 +133,55 @@ impl<const LEN: usize> From<[u8; LEN]> for ByteSet {
...
+                3 => [0, 0, 0, mask],
+                _ => unreachable!("u8 / 64 not in range 0..=3 ??"),
+            };
+            ByteSet { bits }
+        } else {
+            // Fall-back to FromIter implementation, which is less efficient in general
+            value.collect::<ByteSet>()
+        }
+    }
+}
```
mikeday, line 177: I think this is overkill; ranged byteset construction is not in the fast path. :sweat_smile:

src/byte_set.rs
archaephyrryx: The const keyword is enough, and the compiler might inline even without the attribute. I can remove it if it is excessive.

src/bin/doodle/format/base.rs
archaephyrryx: Doesn't ASCII imply 7-bit ASCII?

man ascii(7)
> ASCII is the American Standard Code for Information Interchange.  It is a 7-bit code.

src/byte_set.rs
archaephyrryx: Indeed, there needs to be an edge-case to handle such ranges.

src/byte_set.rs
archaephyrryx: I use it a lot for UTF-8.

src/byte_set.rs
archaephyrryx: Removed, it can always be added back if appropriate.

src/byte_set.rs
archaephyrryx: Fixed this in the latest commit.

src/byte_set.rs
```diff
@@ -126,21 +132,64 @@ impl<const LEN: usize> From<[u8; LEN]> for ByteSet {
...
         bs
     }
 }
 
-impl From<RangeInclusive<u8>> for ByteSet {
-    fn from(range: RangeInclusive<u8>) -> ByteSet {
-        let mut bs = ByteSet::new();
-        for b in range {
-            bs.insert(b);
+impl From<ops::Range<u8>> for ByteSet {
```
mikeday, line 145: Does this need qualification given that ops is already imported?

src/byte_set.rs
```diff
@@ -126,21 +132,64 @@ impl<const LEN: usize> From<[u8; LEN]> for ByteSet {
...
+    fn from(value: ops::RangeInclusive<u8>) -> Self {
+        // because the values are adjacent, we can optimize if they are within the same quadrant
+        let lo = value.start();
+        let hi = value.end();
+        if hi < lo {
+            return ByteSet::empty();
+        } else if hi == lo {
+            return ByteSet::from([*lo]);
+        }
+        let start_ix = *lo / 64;
```
mikeday, line 165: Would be tempted to use `>> 6` here instead of division.

src/byte_set.rs
```diff
@@ -126,21 +132,64 @@ impl<const LEN: usize> From<[u8; LEN]> for ByteSet {
...
+        let end_ix = *hi / 64;
+        if start_ix == end_ix {
+            let mut mask = 0u64;
+            for i in value {
+                mask |= 1 << i % 64;
```
mikeday, line 170: And `& 63` :nerd_face:

src/byte_set.rs
archaephyrryx: I fix this later on in the utf-8 PR so it would be a difficult merge if I changed it at this point. But point taken.

src/byte_set.rs
archaephyrryx: Done.

src/byte_set.rs
archaephyrryx: Done.

mikeday (approved):
That looks fine.

PR #98, opened 2023-11-06
Speed up gzip decoding by caching dynamic formats

mikeday:


src/lib.rs
```diff
@@ -454,6 +464,8 @@ pub enum Format {
     MatchVariant(Expr, Vec<(Pattern, String, Format)>),
     /// Format generated dynamically
     Dynamic(DynFormat),
+    /// Apply a dynamic format from a named variable in the scope
+    Apply(String),
```
archaephyrryx, line 468: We could use a `Cow<'static, str>` here instead to optimize for cases where the label is a string literal we already know. This pattern can be applied more largely to other `String` signatures and fields in the codebase, but at minimum this is a fairly innocuous swap-in.

mikeday: I think I'd prefer to avoid doing that unless we have reason to believe that the string is being frequently copied.

src/lib.rs
```diff
@@ -715,251 +738,307 @@ enum Decoder {
     Match(Expr, Vec<(Pattern, Decoder)>),
     MatchVariant(Expr, Vec<(Pattern, String, Decoder)>),
     Dynamic(DynFormat),
+    Apply(String),
 }
 
 impl Expr {
-    fn eval(&self, scope: &mut Scope) -> Value {
+    fn eval<'a>(&'a self, scope: &'a mut Scope) -> Cow<'a, Value> {
```
archaephyrryx, line 745: This comment more applies to the branches of the pattern-match in the method definition, but a lot of explicit `Cow::Owned` can be elided in favor of `.into()` on the `Value`. I believe this can also be done for `Borrowed`, but would have to check.

src/lib.rs
```diff
@@ -715,251 +738,307 @@ enum Decoder {
...
 
             Expr::BitAnd(x, y) => match (x.eval_value(scope), y.eval_value(scope)) {
-                (Value::U8(x), Value::U8(y)) => Value::U8(x & y),
-                (Value::U16(x), Value::U16(y)) => Value::U16(x & y),
-                (Value::U32(x), Value::U32(y)) => Value::U32(x & y),
+                (Value::U8(x), Value::U8(y)) => Cow::Owned(Value::U8(x & y)),
+                (Value::U16(x), Value::U16(y)) => Cow::Owned(Value::U16(x & y)),
+                (Value::U32(x), Value::U32(y)) => Cow::Owned(Value::U32(x & y)),
                 (x, y) => panic!("mismatched operands {x:?}, {y:?}"),
             },
```
archaephyrryx, line 801:
```suggestion
            Expr::BitAnd(x, y) => (match (x.eval_value(scope), y.eval_value(scope)) {
                (Value::U8(x), Value::U8(y)) => Value::U8(x & y),
                (Value::U16(x), Value::U16(y)) => Value::U16(x & y),
                (Value::U32(x), Value::U32(y)) => Value::U32(x & y),
                (x, y) => panic!("mismatched operands {x:?}, {y:?}"),
            }).into(),
```

or

```suggestion
            Expr::BitAnd(x, y) => Cow::Owned(match (x.eval_value(scope), y.eval_value(scope)) {
                (Value::U8(x), Value::U8(y)) => Value::U8(x & y),
                (Value::U16(x), Value::U16(y)) => Value::U16(x & y),
                (Value::U32(x), Value::U32(y)) => Value::U32(x & y),
                (x, y) => panic!("mismatched operands {x:?}, {y:?}"),
            }),
```

might be a bit more compact, possibly less legible though...

mikeday: That's good I think!

src/lib.rs
```diff
@@ -715,251 +738,307 @@ enum Decoder {
...
                 }
                 _ => panic!("SubSeq: expected Seq"),
             },
             Expr::FlatMap(expr, seq) => match expr.as_ref() {
-                Expr::Lambda(name, expr) => match seq.eval(scope) {
+                Expr::Lambda(name, expr) => match seq.eval(scope).into_owned() {
                     Value::Seq(values) => {
                         let mut vs = Vec::new();
                         for v in values {
                             scope.push(name.clone(), v);
```
archaephyrryx, line 960: Here is a prime example of where `Cow<'static, str>` would save us allocation costs.

src/lib.rs
mikeday: Done.

src/lib.rs
mikeday: That's a good point, I'll investigate in a future branch.

archaephyrryx (approved):
Looks good.

PR #100, opened 2023-11-06
Recursive formats

mikeday:


src/lib.rs
```diff
@@ -1288,6 +1341,93 @@ impl Format {
         }
         MatchTree::build(module, &fs, Rc::new(Next::Empty)).is_none()
     }
+
+    fn recursion_check(&self, module: &FormatModule, format_ref: FormatRef) -> Result<bool, ()> {
```
archaephyrryx, line 1345: The method name `recursion_check` and the return value do not have an intuitive semantics as far as what each possible return value of `Ok(true)`, `Ok(false)`, and `Err(())` might signify. Either a bespoke enum, a doc-comment, or a rename of the method would make things clearer.

archaephyrryx: A more complex type might be useful than a simple enum if we want to capture more information than ternary classification of the form 'yes'/'no'/'failure'

```suggestion

// Allomorph of Option representing the possibly-unknown/unknowable answer to a question
pub enum Questionable<T> {
    Known(T), // can be resolved consistently without external context or runtime information
    Unknown // cannot be resolved in the absence of runtime information, missing context, or complex logic
}

// For possible branches, alternations, variable or dynamic formats, represents under what conditions a particular property is true
pub enum Consistency {
    Always,
    Never,
    Sometimes, // Definitely not 'Never', strongly suggests not 'Always'
}

pub struct RecursionDiagnostic {
    is_auto_recursive: Questionable<bool>, // indicates whether there is a verbatim reference back to the format in question embedded in its definition
    is_cyclic: Questionable<bool>, // indicates whether any evaluation of the format in question will form a cycle back to itself, even if not directly (e.g. through mutual recursion)
    will_advance_offset: Questionable<Consistency>, // indicates whether a successful decode of the format in question will always result in the offset being advanced by at least one byte
    accepts_end_of_input: Questionable<bool>, // Indicates whether, if there was no input left to consume, whether the format would potentially yield a viable parse
}

/* appropriate methods to combine Diagnostics across branches and terms */
```

PR #104, opened 2023-11-10
Format::IsoUnion

mikeday:
This implements unions where all the branches have the same type, either because they match the same structure (for example three bytes in UTF-8) or because they match records with the same type for the `@value` field. Such unions can return the value directly without the need to wrap it in a variant for each branch.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2722,15 +2722,14 @@ fn make_huffman_codes(lengths: &[usize]) -> Format {
     }
 
     let mut codes = Vec::with_capacity(lengths.len());
-    let mut branches = Vec::new();
 
     for n in 0..lengths.len() {
         let len = lengths[n];
         if len != 0 {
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -2668,29 +2722,23 @@ fn make_huffman_codes(lengths: &[usize]) -> Format {
     }
 
     let mut codes = Vec::with_capacity(lengths.len());
-    let mut branches = Vec::new();
+    let mut branches: Vec<Format> = Vec::new();
```
archaephyrryx, line 2725: `branches` does not appear to be used anymore. Is this intentional? (unused_mut, unused_variables).

archaephyrryx (approved):
Looks good, one unused variable but otherwise good to merge.

src/lib.rs
mikeday: Unintentional, didn't notice the warning due to presence of the other warning :roll_eyes:

PR #106, opened 2023-11-10
Utf8 iso union

archaephyrryx:
Refactor UTF-8 to use the IsoUnion variant introduced in #104.


mikeday:
Looks good, some conflicts to resolve?

mikeday:
I tried fixing the conflict through GitHub interface but it doesn't seem right, I think would be better if you just rebase it on current main that should fix the issue.

PR #108, opened 2023-11-12
Immutable scopes

mikeday:
Create a child scope when you need one and simply discard it when done, this makes scope handling much more robust, without the risk of an early exit leaving the parent scope in an inconsistent state.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -172,7 +172,13 @@ impl Value {
 
     /// Returns `true` if the pattern successfully matches the value, pushing
     /// any values bound by the pattern onto the scope
-    fn matches(&self, scope: &mut Scope<'_>, pattern: &Pattern) -> bool {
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -781,7 +780,7 @@ enum Decoder {
 }
 
 impl Expr {
-    fn eval<'a>(&'a self, scope: &'a mut Scope) -> Cow<'a, Value> {
+    fn eval<'a>(&'a self, scope: &'a Scope<'a>) -> Cow<'a, Value> {
```
archaephyrryx, line 789: If we have enough `&'a Scope<'a>`, it might be worth considering a helper alias `type ScopeRef<'a> = &'a Scope<'a>`.

archaephyrryx (approved):
Looks good to me!

PR #109, opened 2023-11-12
Bringing back Format::Map

mikeday:
This is essentially the same as records with `@value` fields but it is cleaner to make the machinery more explicit instead of depending on magic field names and it's also faster.


src/bin/doodle/format/jpeg.rs
```diff
@@ -353,26 +353,11 @@ pub fn main(module: &mut FormatModule, base: &BaseModule, tiff: &FormatRef) -> F
...
-                            var("v"),
-                        ),
-                        (Pattern::variant("zero", Pattern::Wildcard), Expr::U8(0xFF)),
-                    ],
-                )),
+        Format::IsoUnion(vec![
+            not_byte(0xFF),
+            Format::Map(
+                Box::new(tuple([is_byte(0xFF), is_byte(0x00)])),
+                Expr::Lambda("_".into(), Box::new(Expr::U8(0xFF))),
```
archaephyrryx, line 360: Do we want to have any code to optimize out type-inference, and avoid scope mutation, for the identifier `"_"` in a Lambda preamble?

src/lib.rs
```diff
@@ -672,12 +665,24 @@ impl FormatModule {
             Format::Slice(_expr, a) => self.infer_format_type(scope, a),
             Format::Bits(a) => self.infer_format_type(scope, a),
             Format::WithRelativeOffset(_expr, a) => self.infer_format_type(scope, a),
+            Format::Map(a, expr) => {
```
archaephyrryx, line 668: Do we care at all if the expansion of the lambda is independent of the type of the value it binds? Or do we always want to make sure that type inference on the bound variable resolves even if it is not relevant to the mapped value type?

src/output/tree.rs
```diff
@@ -298,6 +313,13 @@ impl<'module> MonoidalPrinter<'module> {
             Value::Seq(vals) => self.compile_seq(scope, vals, None),
             Value::Record(fields) => self.compile_record(scope, fields, None),
             Value::Variant(label, value) => self.compile_variant(scope, label, value, None),
+            Value::Mapped(orig, value) => {
+                if self.flags.collapse_mapped_values {
+                    self.compile_value(scope, value)
+                } else {
+                    self.compile_value(scope, orig)
```
archaephyrryx, line 320: Is it proper to show only the original, and not the mapped value, if we have the `collapse_mapped_values` flag set? Or does this happen elsewhere?

tests/expected/decode/test.tar.stdout
```diff
@@ -27,19 +27,19 @@
...
-│           │       ├── 2 := 22
-│           │       ├── 3 := 7
-│           │       ├── 4 := 0
-│           │       ├── 5 := 2
-│           │       ├── 6 := 0
-│           │       ├── 7 := 0
-│           │       ├── 8 := 77
-│           │       ├── 9 := 97
+│           │   └── file <- repeat-count (header.size) base.u8 :=
+│           │       ├── 0 <- base.u8 := 0
```
archaephyrryx, line 31: Because this is a `repeat-count`, we already know what the type of each index will be. How do we rank the quality of this changed output, compared to before (even if it is more of a side-effect than a conscious decision)?

tests/expected/decode/test4.gz.stdout
```diff
@@ -13,24 +13,24 @@
 │           │   ├── blocks <- repeat-until-last (x -> x.final == 1) deflate.block :=
 │           │   │   ├── 0 <- deflate.block :=
 │           │   │   │   ├── final <- base.bit := 0
-│           │   │   │   ├── type <- { ... } := 2
+│           │   │   │   ├── type <- map (bits -> bits.1 << 1 | bits.0) (...) := 2
```
archaephyrryx, line 16: It might be beneficial to have 'named lambdas' that display under a simple operation name, rather than force it to expand potentially verbose expressions into text. We could even add a threshold level for how many characters the expansion is allowed to take up before it falls back to the name, but that might be over-engineering an already somewhat extraneous feature...

archaephyrryx (approved):

src/bin/doodle/format/jpeg.rs
mikeday: I think this will be easier to do with a future compilation scheme that breaks expressions into simpler components instead of just walking the AST.

src/lib.rs
mikeday: We still want to perform type inference on the format being mapped, even if the value is ignored.

src/output/tree.rs
mikeday: No I think we should show both the original and mapped values and currently it only shows the original, this is a limitation. I'm not sure what's the best way to show the original value, like how we should present both values, perhaps it could be on two lines? With names "orig" and "mapped" or something? It's a bit clunky...

tests/expected/decode/test.tar.stdout
mikeday: Ah good point, it's a side-effect but might be worth checking to see if we can do better.

tests/expected/decode/test4.gz.stdout
mikeday: Yes we're a bit limited by the nature of a basic ASCII text dump here, but you could imagine that large values would be summarised or collapsed until explicitly expanded, in a more powerful UI.

src/output/tree.rs
archaephyrryx: We could have something like

Inline Version

```
<fname> <- map (lambda) := <orig> ↦ <value>
```

Multiline Version

```
<fname> <- map (lambda) := <orig>
     ↳<value>
```

mikeday: Yes, although although most common is for orig to be multiline and value to be inline, for example base.u16 taking a tuple of two bytes to one U16.

tests/expected/decode/test4.gz.stdout
archaephyrryx: It might be overkill or a diversion, but the [tui crate](https://crates.io/crates/tui) looks pretty neat. It would be interesting, if we decided it was worth the effort, to see what we could do with a more feature-full interaction between the user and the output of the program.

PR #110, opened 2023-11-12
Value::Branch, to remember the branch taken in union and match formats

mikeday:
This value is semantically transparent and only exists to trace the path of execution that created it, however it may be necessary for display purposes, to ensure we know the correct format.

### DIFF
--- a/src/output/flat.rs
+++ b/src/output/flat.rs
@@ -93,8 +93,16 @@ fn is_show_format(name: &str) -> Option<&'static str> {
         "jpeg.scan-data" => Some("Entropy-Coded Segment"),
 
         // Tar
+        "tar.main" => Some("tar.main"),
         "tar.header" => Some("Tar Header"),
         "tar.header_with_data" => Some("Tar File Entry"),
+
+        // Text
+        "text" => Some("ASCII text"),
+
+        // UTF-8
+        "utf8.string" => Some("UTF-8"),
+
         _ => None,
     }
 }
@@ -165,7 +173,7 @@ fn check_covered(
             check_covered(module, path, format)?;
         }
         Format::WithRelativeOffset(_, _) => {} // FIXME
-        Format::Map(_format, _expr) => {}
+        Format::Map(format, _expr) => check_covered(module, path, format)?,
         Format::Compute(_expr) => {}
         Format::Match(_head, branches) => {
             for (_pattern, format) in branches {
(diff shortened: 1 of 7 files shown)


archaephyrryx (approved):

src/output/flat.rs
```diff
@@ -217,14 +217,23 @@ impl<'module, W: io::Write> Context<'module, W> {
                 }
                 _ => panic!("expected variant"),
             },
-            Format::Union(branches) | Format::NondetUnion(branches) => match value {
-                Value::Variant(label, value) => {
-                    let (_, format) = branches.iter().find(|(l, _)| l == label).unwrap();
+            Format::UnionVariant(branches) | Format::UnionNondet(branches) => match value {
+                Value::Branch(index, value) => {
+                    let format = &branches[*index].1;
```
archaephyrryx, line 230: Do we potentially want a sanity-check that `&branches[*index].0 == _label`, or is that perfunctory?

src/output/tree.rs
```diff
@@ -123,11 +123,8 @@ impl<'module> MonoidalPrinter<'module> {
             Value::Tuple(values) => values.is_empty(),
             Value::Record(fields) => fields.is_empty(),
             Value::Seq(values) => values.is_empty(),
-            Value::Variant(label, value) => match format {
-                Some(Format::Union(branches)) | Some(Format::NondetUnion(branches)) => {
-                    let (_, format) = branches.iter().find(|(l, _)| l == label).unwrap();
-                    self.is_atomic_value(value, Some(format))
-                }
+            Value::Variant(_label, value) => match format {
+                Some(Format::Variant(_label, format)) => self.is_atomic_value(value, Some(format)),
```
archaephyrryx, line 127: Same here, do we want to sanity-check `_label/0 == _label/1`

src/output/tree.rs
```diff
@@ -142,6 +139,17 @@ impl<'module> MonoidalPrinter<'module> {
...
+            Value::Branch(n, value) => match format {
+                Some(Format::Union(branches)) => {
+                    let format = &branches[*n];
+                    self.is_atomic_value(value, Some(format))
+                }
+                Some(Format::UnionVariant(branches)) | Some(Format::UnionNondet(branches)) => {
+                    let format = &branches[*n].1;
+                    self.is_atomic_value(value, Some(format))
+                }
+                _ => self.is_atomic_value(value, None),
```
archaephyrryx, line 151: Do we care about the difference between seeing `None` vs. `Some(_x)` here, where `_x` is implicitly a non-union-oriented format? In other words, is an unexpected format in this position deemed a panic-worthy error, or is it sensible to gloss over those cases?

src/output/tree.rs
```diff
@@ -270,30 +289,34 @@ impl<'module> MonoidalPrinter<'module> {
...
                         return frag;
                     }
+                    panic!("pattern match failure");
                 }
-                panic!("non-exhaustive patterns");
-            }
-            Format::MatchVariant(head, branches) => {
-                let head = head.eval(&scope);
-                for (pattern, _label, format) in branches {
+                _ => panic!("expected branch value"),
```
archaephyrryx, line 302: We should possibly keep up the pattern of `other => panic!("expected branch value, found {other:?}")` to ensure that the panics, though rare, give us a reasonable amount of information to work with.

src/output/tree.rs
```diff
@@ -270,30 +289,34 @@ impl<'module> MonoidalPrinter<'module> {
...
                         } else {
                             panic!("expected variant value");
                         }
                         return frag;
                     }
+                    panic!("pattern match failure");
                 }
-                panic!("non-exhaustive patterns");
-            }
+                _ => panic!("expected branch value"),
```
archaephyrryx, line 318: Same as above, `other => panic!("<...> {other:?}")` might be preferable?

mikeday:
I think all issues have been addressed now.

PR #111, opened 2023-11-12
Introduce Label alias for Cow<'static, str> in lib

archaephyrryx:
Many sites within `lib.rs` and modules that extend or rely on related API details currently have `Cow<'static, str>` as a stand-in for `String`. This PR simply aliases the majority of such instances that share the same semantic intent as `etc::Label`, where `Label` is a simple type-alias defined in a new `etc.rs` submodule of `lib.rs`.

This is more of a stylistic change than anything and can easily be skipped if there is there is no compelling reason to adopt this alias.

### DIFF
--- a/src/bin/doodle/format/base.rs
+++ b/src/bin/doodle/format/base.rs
@@ -1,13 +1,12 @@
 use doodle::byte_set::ByteSet;
 use doodle::{Expr, Format, FormatModule, FormatRef, Pattern};
+use doodle::etc;
 
-use std::borrow::Cow;
-
-pub fn var<Name: Into<Cow<'static, str>>>(name: Name) -> Expr {
+pub fn var<Name: Into<etc::Label>>(name: Name) -> Expr {
     Expr::Var(name.into())
 }
 
-pub fn bind<Name: Into<Cow<'static, str>>>(name: Name) -> Pattern {
+pub fn bind<Name: Into<etc::Label>>(name: Name) -> Pattern {
     Pattern::Binding(name.into())
 }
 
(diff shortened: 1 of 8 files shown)


mikeday:
Could we just define Label in lib instead of defining an etc module just for that?

archaephyrryx:
> Could we just define Label in lib instead of defining an etc module just for that?

That wouldn't normally be an issue, but we currently have a couple of methods/functions declared as `fn foo<Label: Into<etc::Label>>`, where having a partially qualified name rather than a direct import saves us on more complex refactoring. We can always move it later, but for now it is easier to effect the change in this way.

mikeday:
I don't quite understand the comment about a partially qualified name saving us on more complex refactoring, is it because they use "Label" as a type variable? Could we change it to `LabelType: Info<Label>`?

archaephyrryx:
> I don't quite understand the comment about a partially qualified name saving us on more complex refactoring, is it because they use "Label" as a type variable? Could we change it to `LabelType: Into<Label>`?

Yes, and I am working on refactoring so that the alias can be used unqualified. At this point it should be straightforward enough to collapse into `lib.rs` in one more commit.

mikeday:
That would be ideal I think, as any code using any of these types is invariably going to use Label as well.

archaephyrryx:
`etc.rs` has been pruned, and its definitions are now top-level items in `lib.rs`. For brevity, the generic `Name` is used wherever the verbatim identifier `Label` appeared before. We also have an `IntoLabel` trait that could be discarded, as it is just a marker trait for `Into<Label>` and saves us only 2 keystrokes out of 11. On the other hand, it can make editor autocomplete for in-crate identifiers a bit easier when we want `Into<Label>`. Either way, even with just the alias `Label`, this is more or less where the PR ought to be at this point.

mikeday (approved):
Looks good to me!

PR #112, opened 2023-11-13
Omit redundant types and formats for Seq

archaephyrryx:
Adds an extra argument to `compile_field_value_{continue,last}` that determines whether explicit format information is necessary based on context available to the caller (`true` when called from `compile_record` or `compile_tuple`, `false` when called from `compile_seq`). Also adds an override flag that, if set, will invariably show the format across items in a repetition.

Updates expected test output accordingly.


tests/expected/decode/test.gif.stdout
```diff
@@ -45,7 +45,7 @@
 │       │              0   0   0
 │       │              0   0   0
 │       ├── blocks <- repeat gif.block :=
-│       │   └── 0 <- gif.block :=
+│       │   └── 0 :=
```
mikeday, line 48: Losing the format here seems mildly disadvantageous, would it be possible to keep it for the multiline case, given that the rest of the line is free anyway, and only omit it for the atomic value case?

tests/expected/decode/test6.gz.stdout
```diff
@@ -11,26 +11,26 @@
...
+│           │   │           │   ├── 0 := 13
+│           │   │           │   ├── 1 := 60
+│           │   │           │   ├── 2 := 105
+│           │   │           │   ├── 3 := 200
+│           │   │           │   ├── 4 := 176
+│           │   │           │   ├── 5 := 100
+│           │   │           │   ├── 6 := 108
+│           │   │           │   ├── 7 := 167
+│           │   │           │   ├── 8 := 57
+│           │   │           │   ├── 9 := 17
```
mikeday, line 31: Certainly a win here :laughing:

tests/expected/decode/test.gif.stdout
archaephyrryx: I added a perhaps-precarious condition under which the omission will not occur (format is ItemVar, Apply, or Dynamic, and the value-fragment spans multiple lines). It seems to do the trick, but might be a bit messy to maintain or have edge-cases we might not yet have coverage for.

src/output/tree.rs
```diff
@@ -611,22 +616,37 @@ impl<'module> MonoidalPrinter<'module> {
...
-            frags.push(Fragment::String(" <- ".into()));
-            frags.push(self.compile_format(format, Precedence::FORMAT_COMPOUND));
+            let tmp = self.compile_field_value(scope, value, Some(format));
+            if (self.is_indirect_format(format) && !tmp.is_single_line(true)) || format_needed || self.flags.show_redundant_formats {
+                frags.push(Fragment::String(" <- ".into()));
+                frags.push(self.compile_format(format, Precedence::FORMAT_COMPOUND));
+            }
         }
         self.gutter.push(Column::Branch);
         frags.push(self.compile_field_value(scope, value, format));
```
mikeday, line 652: Might be able to refactor this to avoid calling `compile_field_value` twice with the same arguments.

src/output/tree.rs
```diff
@@ -640,16 +660,22 @@ impl<'module> MonoidalPrinter<'module> {
...
+
+        if let Some(format) = format  {
+            let tmp = self.compile_field_value(scope, value, Some(format));
+            if (self.is_indirect_format(format) && !tmp.is_single_line(true)) || format_needed || self.flags.show_redundant_formats {
+                frags.push(Fragment::String(" <- ".into()));
+                frags.push(self.compile_format(format, Default::default()));
+            }
         }
         self.gutter.push(Column::Space);
         frags.push(self.compile_field_value(scope, value, format));
```
mikeday, line 681: Similarly here.

src/output/tree.rs
archaephyrryx: Working on it right now. Luckily we are already monoidal so the fragment generation can happen out-of-order very easily.

archaephyrryx: Done!

src/output/tree.rs
archaephyrryx: Handled as well.

mikeday (approved):

PR #114, opened 2023-11-13
General improvements to Fragment type

archaephyrryx:
Adds methods to `output::Fragment`, with the expectation that they might soon be useful, even if not currently used anywhere.

Improves documentation of Fragment type.

Minor fix to comments on `tree::Relation` enum variants.


mikeday (approved):
That looks fine.

PR #115, opened 2023-11-14
Unify utf-8 and text into text

archaephyrryx:
Addresses issue #105 by merging `utf8.rs` and `text.rs` into a new format module `text.rs`

Defines `text` as a non-deterministic union over `text.ascii` (:= old `text`) and `text.utf8` (:= old `utf8`). This may be overkill, as `utf8` itself contains ascii as a strict subset.

Updates `tree.rs` predicates for treating formats as utf-8 strings, and adds a match-case that allows 'ascii-string' values to be raw sequences, rather than enforcing that they must always be records.


mikeday:
This looks fine but has conflicts and can you try running `cargo run file --output flat test.utf8` and fix `is_show_format` in output/flat appropriately.

archaephyrryx:
> This looks fine but has conflicts and can you try running `cargo run file --output flat test.utf8` and fix `is_show_format` in output/flat appropriately.

What are the semantics of  `is_show_format`?

mikeday:
It specifies whether a format should be included in flat output or not, and if so what it should be titled; the flat output is used to provide a linear "table of contents" for convenient indexing into a file dump. Given that a text file has no sections or structure and we don't want an item for each character, it makes sense to just make the top-level format a show format.

(The way flat output works is that every format capable of consuming bytes must either be a show format or contained within a show format; this works out very nicely for PNG and JPEG but less well for some hierarchical formats like RIFF, which are difficult to flatten).

archaephyrryx:
> It specifies whether a format should be included in flat output or not, and if so what it should be titled; the flat output is used to provide a linear "table of contents" for convenient indexing into a file dump. Given that a text file has no sections or structure and we don't want an item for each character, it makes sense to just make the top-level format a show format.
> 
> (The way flat output works is that every format capable of consuming bytes must either be a show format or contained within a show format; this works out very nicely for PNG and JPEG but less well for some hierarchical formats like RIFF, which are difficult to flatten).

Adopted the necessary changes while rebasing onto the updated main.

src/output/flat.rs
```diff
@@ -98,10 +98,8 @@ fn is_show_format(name: &str) -> Option<&'static str> {
         "tar.header_with_data" => Some("Tar File Entry"),
 
         // Text
-        "text" => Some("ASCII text"),
-
-        // UTF-8
-        "utf8.string" => Some("UTF-8"),
+        "text.string.ascii" => Some("ASCII String"),
+        "text.string.utf8" => Some("UTF-8 String"),
```
mikeday, line 102: Can probably say "ASCII Text" and "UTF-8 Text" I think, given that these are a little more friendly in this context.

mikeday (approved):
Looks good!

PR #116, opened 2023-11-14
Subset alternation

archaephyrryx:
Adds a `RepeatFallback` and related helper-variants to Decoder, Format, and Value.

$\mathrm{RepeatFallback}(A, B) \coloneqq \mathrm{Repeat}(A) \cup \mathrm{PeekNot}(\mathrm{Repeat}(A)) \oplus \mathrm{Repeat}(B)$

There is a strong possibility this is misimplemented somewhere due to knock-on effect of failure-tracing to different sites, which may have left things in a partially-implemented or hold-over state.


archaephyrryx:
Note that this changeset treats #115 as its upstream, so it will be hard to merge until #115 is merged.

src/lib.rs
```diff
@@ -2551,6 +2604,44 @@ impl Decoder {
...
+                    match superset.parse(program, scope, incr_input) {
+                        Ok((b, next_input)) => {
+                            accum.push(b);
+                            incr_input = next_input;
+                        }
+                        Err(_err) => break,
+                    }
+                }
+
+                if let Err(e) = (Decoder::EndOfInput).parse(program, scope, incr_input) {
```
archaephyrryx, line 2634: This is obviously wrong, but the MatchTree logic for how to judge effective end-of-RepeatFallback eludes me...

src/output/flat.rs
```diff
@@ -278,6 +282,15 @@ impl<'module, W: io::Write> Context<'module, W> {
                 }
                 _ => panic!("expected sequence, found {value:?}"),
             },
+            Format::RepeatFallback(_narrow, _wide) => match value {
+                Value::Seq(values) => {
+                    for _v in values {
+                        (); // FIXME
```
archaephyrryx, line 288: I am slightly unclear on the semantics of `check_covered`, and so I am not sure how to implement this...

PR #118, opened 2023-11-15
Take Decoders out of Scope and put them Value instead.

mikeday:
This simplifies Scope at the cost of complicating Value; we don't really want the cached Decoder object to interfere with comparing, hashing, debug printing, or JSON serialising Value!

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -2,14 +2,12 @@ use crate::byte_set::ByteSet;
 use crate::error::{ParseError, ParseResult};
 use crate::read::ReadCtxt;
 use crate::{DynFormat, Expr, Format, FormatModule, MatchTree, Next, Pattern, ValueType};
-use serde::Serialize;
 use std::borrow::Cow;
 use std::cell::RefCell;
 use std::collections::HashMap;
 use std::rc::Rc;
 
-#[derive(Clone, PartialEq, Eq, Hash, Debug, Serialize)]
-#[serde(tag = "tag", content = "data")]
+#[derive(Clone, PartialEq, Debug)]
 pub enum Value {
     Bool(bool),
     U8(u8),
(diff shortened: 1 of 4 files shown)


archaephyrryx:
How do we rank the relative merits of

(A)

```rust
pub enum Value {
    /* ... */
    Format(Box<Format>, RefCell<Option<Decoder>>),
}

/* ... */

pub struct Scope<'a> {
     parent: Option<&'a Scope<'a>>,
     names: Vec<Cow<'static, str>>,
     values: Vec<Value>
}
```

and

(B) 
```rust
pub enum ScopeEntry {
     Value(Value),
     Format(Format, RefCell<Option<Decoder>>),
}

pub struct Scope<'a> {
    parent: Option<&'a Scope<'a>>,
    names: Vec<Cow<'static, str>>,
    entries: Vec<ScopeEntry>,
}
```

I don't see a problem with either, but I don't quite know which is better...

mikeday:
Well the `ScopeEntry` approach seems like a strict improvement on having a separate decoders vec anyway!

src/decoder.rs
```diff
@@ -22,7 +20,7 @@ pub enum Value {
     Seq(Vec<Value>),
     Mapped(Box<Value>, Box<Value>),
     Branch(usize, Box<Value>),
-    Format(Box<Format>),
+    Format(Box<Format>, RefCell<Option<Decoder>>),
```
archaephyrryx, line 23: We can use `std::cell:OnceCell` instead, since this is a cached value that is initialized once and never mutated after.

archaephyrryx:
```suggestion
Format(Box<Format>, OnceCell<Decoder>),
```

mikeday:
Closing this as it's nasty.

PR #119, opened 2023-11-15
Make TypeScope immutable similar to Scope.

mikeday:


archaephyrryx:
The parallelism strongly suggests that there should be some version of a `GenericScope<'a, T>` struct somewhere, with `type TypeScope<'a> = GenericScope<'a, Type>` and `type Scope<'a> = GenericScope<'a, Value>`. We can always do that later, just a thought though.

PR #120, opened 2023-11-15
Attempted implementation of add for WithRelativeOffset

archaephyrryx:


src/lib.rs
```diff
@@ -1247,8 +1247,26 @@ impl<'a> MatchTreeStep<'a> {
             Format::Bits(_a) => {
                 Self::accept() // FIXME
             }
-            Format::WithRelativeOffset(_expr, _a) => {
-                Self::accept() // FIXME
+            Format::WithRelativeOffset(expr, a) => {
+                // REVIEW - this is a hack but it is at least somewhat better than before
+                let tree = Self::add_next(module, next.clone());
+                let bounds = expr.bounds();
+                let peek = match bounds.is_exact() {
```
mikeday, line 1254: I think we are better off skipping the peek entirely and just doing the add_next part.

archaephyrryx: Sorry, what specifically would the code look like after that proposed change? Not quite sure what your suggestion entails...

mikeday: I think that checking for the slice isn't helpful, as it's just an unknown number of unknown bytes, not something a match tree can lookahead to find. This part is useful:
```rust
                let tree = Self::add_next(module, next.clone());
```
As this is ignoring the offset completely and just moving on to the next format, which would handle the case of a file that starts with an offset to something followed by a fixed signature, an actual possibility.

src/lib.rs
```diff
@@ -1247,8 +1247,26 @@ impl<'a> MatchTreeStep<'a> {
...
+                        n,
+                        Rc::new(Next::Empty),
+                        Rc::new(Next::Tuple(std::slice::from_ref(a.as_ref()), next)),
+                    ),
+                    None => Self::add_slice(
+                        module,
+                        bounds.min,
+                        Rc::new(Next::Empty),
+                        Rc::new(Next::Empty),
+                    ),
```
mikeday, line 1267: Or at least this part is unlikely to be necessary.

src/lib.rs
archaephyrryx: I guess the question is, are any byte-level restrictions on the WithRelativeOffset scrutinee salient enough, that they might make-or-break an input's eligibility to parse as a certain format? Since WithRelativeOffset is used rarely enough, and usually in contexts where there is no viable alternate format to parse the input as, at the point in the input where the WithRelativeOffset format would be encountered, that the MatchTree can safely disregard any issues with the data in question being ambiguous with respect to another format. And therefore, that the bytes being scrutinized by the offset-peek can be virtually anything, and the actual attempt to compile can catch any wrongdoing at the byte-level. In theory the MatchTree should perhaps care about the bytes being inspected, but without unbounded lookahead, that can't be done universally I suppose, and in any case, we aren't at the point where such considerations matter; it probably won't be too late to reimplement a more-correct version later on, if we run into any branch-ambiguities between formats where the WithRelativeOffset format would be able to tie-break if fully processed...

src/lib.rs
archaephyrryx: Ah, I see. I can narrow the peek to just the cncrete case, and otherwise just default to the value `tree`. Does that work?

archaephyrryx: I changed it accordingly, let me know if there are still any points to revise.

mikeday (approved):
I think that's fine.

PR #121, opened 2023-11-15
Handle dynamic formats with Let bindings

mikeday:
The way this works is that there is a new `Format::Let(name, expr, format)` construct for let bindings which can only be used to bind format values and is the _only_ way to bind format values; dynamic formats cannot be record fields any more.

With these invariants in place, we can take decoders out of the scope and place them on a separate "dynstack" at runtime which only contains decoders created by format let bindings and thus can be directly indexed by integer offset.

Whether this is an improvement or not requires further consideration; it definitely simplifies scopes, but at the cost of introducing a new mutable scope-like thing. On the other hand, it doesn't need RefCell.

Edit: it's also unsound in the presence of pattern matching, which could pull a format out of a computed tuple field.

### DIFF
--- a/src/bin/doodle/format/deflate.rs
+++ b/src/bin/doodle/format/deflate.rs
@@ -239,179 +239,179 @@ pub fn main(module: &mut FormatModule, base: &BaseModule) -> FormatRef {
         "deflate.fixed_huffman",
         record([
             (
-                "format",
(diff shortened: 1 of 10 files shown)


archaephyrryx:
Could this model be used to implement recursive formats? Not sure if that would overcomplicate things, but auto-referential let bindings wouldn't be too hard to implement; on the other hand, they might be nightmarish to write the logic for to detect, sanity-check, and handle appropriately...

PR #122, opened 2023-11-15
ScopeIter changes

mikeday:
This changes ScopeIter so it doesn't clone the scope it iterates over and also reverses the direction of iteration and recurses into parent scopes, so innermost bindings will be returned first.

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -540,13 +540,13 @@ pub struct Scope<'a> {
     decoders: Vec<RefCell<Option<Decoder>>>,
 }
 
-pub struct ScopeIter {
-    name_iter: std::vec::IntoIter<Cow<'static, str>>,
-    value_iter: std::vec::IntoIter<Value>,
+pub struct ScopeIter<'a> {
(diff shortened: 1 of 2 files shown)


src/decoder.rs
```diff
@@ -595,7 +603,7 @@ impl<'a> Scope<'a> {
         }
     }
 
-    pub fn iter(&self) -> impl Iterator<Item = (Cow<'static, str>, Value)> {
+    pub fn iter(&'a self) -> impl Iterator<Item = (&'a Cow<'static, str>, &'a Value)> {
```
archaephyrryx, line 606: I believe that lifetlime elision can be used to avoid the need for `'a` here:

```suggestion
    pub fn iter(&self) -> impl Iterator<Item = (&Cow<'static, str>, &Value)> {
```

But I would have to double-check that this would work...

archaephyrryx:
Looks good to me!

archaephyrryx (approved):

src/decoder.rs
mikeday: Thanks, fixed.

PR #123, opened 2023-11-16
Refactoring, documentation, and clippy-informed cleanup

archaephyrryx:
Initial attempt at documenting undocumented items in the `MatchTree*` family, along with some minor reordering of items and blocks in `lib.rs`.

Also includes a number of disparate changes based on `clippy` suggestions.

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -371,7 +371,7 @@ impl Expr {
                         let ret = expr.eval_lambda(scope, Value::Tuple(vec![accum, v]));
                         accum = match ret.unwrap_tuple().as_mut_slice() {
                             [accum, Value::Seq(vn)] => {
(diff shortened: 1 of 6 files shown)


src/decoder.rs
```diff
@@ -901,7 +900,7 @@ impl Decoder {
             Format::PeekNot(a) => {
                 const MAX_LOOKAHEAD: usize = 1024;
                 match a.match_bounds(compiler.module).max {
-                    None => return Err(format!("PeekNot cannot require unbounded lookahead")),
+                    None => return Err("PeekNot cannot require unbounded lookahead".to_string()),
```
mikeday, line 903: I'd rather stick with format! here as it's shorter and easier to add args when debugging

mikeday (approved):
That looks fine.

PR #126, opened 2023-11-16
Use existing type-alias to simplify signatures

archaephyrryx:
Performs a very easy, but overlooked refactoring in `tree.rs`


mikeday (approved):
Looks good!

PR #127, opened 2023-11-17
Let bindings, separate kinds for values and formats

mikeday:
This improves on #121 by adding as simple kind system that separates format types from value types.

Expression evaluation can only return a value type, not a format type. In particular variables can only evaluate to values, not formats. This makes it impossible to put a format in a record or a tuple or pattern match on formats or anything like that.

Let bindings work for value types, `let name := expr in format` behaves as you would expect.

Dynamic bindings compile a dynamic format: `dynamic name := format in format` will compile the specified dynamic format to a decoder that can be applied by name.

Compiled decoders live in the scope, stored in a ScopeEntry that is either a value or a decoder.

As far as I can tell this scheme is sound and seems a lot easier to prove that it's sound.

### DIFF
--- a/src/output/tree.rs
+++ b/src/output/tree.rs
@@ -325,6 +325,12 @@ impl<'module> MonoidalPrinter<'module> {
                 }
             }
             Format::Compute(_expr) => self.compile_value(scope, value),
+            Format::Let(name, expr, format) => {
+                let v = expr.eval_value(&scope);
+                let mut let_scope = Scope::child(scope);
+                let_scope.push(name.clone(), v);
+                self.compile_decoded_value(&let_scope, value, format)
+            }
             Format::Match(head, branches) => match value {
                 Value::Branch(index, value) => {
                     let head = head.eval(scope);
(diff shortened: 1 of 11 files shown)


src/output/tree.rs
```diff
@@ -354,7 +360,11 @@ impl<'module> MonoidalPrinter<'module> {
                 }
                 _ => panic!("expected branch, found {value:?}"),
             },
-            Format::Dynamic(_) => self.compile_value(scope, value),
+            Format::Dynamic(name, _dynformat, format) => {
+                let mut child_scope = Scope::child(scope);
+                child_scope.push(name.clone(), Value::Tuple(vec![])); // FIXME?
```
archaephyrryx, line 365: Do we have an idea of what should go here, or what we might need to implement to have an appropriate replacement for this scaffolding?

archaephyrryx (approved):
I think it is definitely an improvement to not have to add pseudo-fields for Dynamic formats to format-records, just to be refer back to a scoped item later on.

src/output/tree.rs
mikeday: For now I don't think we need anything, in future we could potentially save the generated format for optional display, I will update the comment to say that.

PR #129, opened 2023-11-19
Multiple Scope types for efficiency and convenience.

mikeday:
This halves the runtime for test4.gz.

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -572,7 +572,7 @@ impl<'a> Scope<'a> {
     fn get_value_by_name(&self, name: &str) -> &Value {
         match self {
             Scope::Empty => panic!("value not found: {name}"),
-            Scope::Multi(normal) => normal.get_value_by_name(name),
+            Scope::Multi(multi) => multi.get_value_by_name(name),
             Scope::Single(single) => single.get_value_by_name(name),
             Scope::Decoder(decoder) => decoder.parent.get_value_by_name(name),
         }
@@ -581,7 +581,7 @@ impl<'a> Scope<'a> {
     fn get_decoder_by_name(&self, name: &str) -> &Decoder {
         match self {
             Scope::Empty => panic!("decoder not found: {name}"),
-            Scope::Multi(normal) => normal.parent.get_decoder_by_name(name),
+            Scope::Multi(multi) => multi.parent.get_decoder_by_name(name),
             Scope::Single(single) => single.parent.get_decoder_by_name(name),
             Scope::Decoder(decoder) => decoder.get_decoder_by_name(name),
         }
@@ -590,7 +590,7 @@ impl<'a> Scope<'a> {
     pub fn get_bindings(&self, bindings: &mut Vec<(Label, ScopeEntry)>) {
         match self {
             Scope::Empty => {}
-            Scope::Multi(normal) => normal.get_bindings(bindings),
+            Scope::Multi(multi) => multi.get_bindings(bindings),
             Scope::Single(single) => single.get_bindings(bindings),
             Scope::Decoder(decoder) => decoder.get_bindings(bindings),
         }
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -1126,10 +1126,9 @@ impl Decoder {
                     v.push(va);
                     let vs = Value::Seq(v);
                     let done = expr.eval_lambda(scope, &vs).unwrap_bool();
-                    v = if let Value::Seq(v) = vs {
-                        v
-                    } else {
-                        panic!("wut");
+                    v = match vs {
+                        Value::Seq(v) => v,
+                        _ => unreachable!(),
                     };
                     if done {
                         break;
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -404,7 +404,7 @@ impl Expr {
     fn eval_lambda<'a>(&self, scope: &'a Scope<'a>, arg: &Value) -> Value {
         match self {
             Expr::Lambda(name, expr) => {
-                let child_scope = SingleScope::new(scope, name.clone(), arg);
+                let child_scope = SingleScope::new(scope, name, arg);
                 expr.eval_value(&Scope::Single(child_scope))
             }
             _ => panic!("expected Lambda"),
@@ -558,13 +558,13 @@ pub struct NormalScope<'a> {
 
 pub struct SingleScope<'a> {
     parent: &'a Scope<'a>,
-    name: Label,
+    name: &'a str,
     value: &'a Value,
 }
 
 pub struct DecoderScope<'a> {
     parent: &'a Scope<'a>,
-    name: Label,
+    name: &'a str,
     decoder: Decoder,
 }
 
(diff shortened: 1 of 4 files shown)


src/decoder.rs
```diff
@@ -1100,8 +1124,13 @@ impl Decoder {
...
                     v.push(va);
-                    let vs = Value::Seq(v.clone());
-                    let done = expr.eval_lambda(scope, vs).unwrap_bool();
+                    let vs = Value::Seq(v);
+                    let done = expr.eval_lambda(scope, &vs).unwrap_bool();
+                    v = if let Value::Seq(v) = vs {
+                        v
+                    } else {
+                        panic!("wut");
+                    };
```
mikeday, line 1133: Is there a nicer way of doing this?

archaephyrryx: Yes!

```suggestion
let vs = Value::Seq(v);
let done = expr.eval_lambda(scope, &vs).unwrap_bool();
let Value::Seq(v) = vs else { unreachable!("wut") };
```

or even

```suggestion
let done = expr.eval_lambda(scope, &Value::Seq(v)).unwrap_bool();
```

Not sure if the second will complain about lifetimes but the first definitely should work.

archaephyrryx:
I would suggest a name like `MultiScope` or `VecScope` over `NormalScope`, as there is no real significance of 'normal' is, outside of what Scope used to be defined as; and that would no longer be transparent after this change is effected.

src/decoder.rs
```diff
@@ -546,117 +545,143 @@ pub enum ScopeEntry {
...
-                    *self = parent.into_iter();
-                    self.next()
-                }
-                None => None,
-            },
-        }
-    }
+pub struct SingleScope<'a> {
+    parent: &'a Scope<'a>,
+    name: Cow<'static, str>,
```
archaephyrryx, line 562: Is it possible that we can replace this with `&'a str` for cheaper Copy than Clone? We already know that `SingleScope<'a>` is bound by the lifetime constraint `: 'a`, based on the other two elements...

src/decoder.rs
```diff
@@ -546,117 +545,143 @@ pub enum ScopeEntry {
...
+    value: &'a Value,
 }
 
-impl<'a> IntoIterator for &'a Scope<'a> {
-    type Item = (&'a Cow<'static, str>, &'a ScopeEntry);
+pub struct DecoderScope<'a> {
+    parent: &'a Scope<'a>,
+    name: Cow<'static, str>,
```
archaephyrryx, line 568: Same idea as with `SingleScope<'a>`, because this is a short-lived stack allocation, we don't need a 'static/owned value, just a reference that lives at least as long as `'a`.

src/decoder.rs
```diff
@@ -546,117 +545,143 @@ pub enum ScopeEntry {
     Decoder(Decoder),
 }
 
-pub struct Scope<'a> {
-    parent: Option<&'a Scope<'a>>,
-    names: Vec<Cow<'static, str>>,
-    entries: Vec<ScopeEntry>,
+pub enum Scope<'a> {
+    Empty,
+    Normal(&'a NormalScope<'a>),
```
archaephyrryx, line 550: I think it would be better to have `Vec(&'a VecScope<'a>)` or `Multi(&'a MultiScope<'a>)`, as `Normal` doesn't quite explain the trichotomy clearly enough in my view.

src/decoder.rs
archaephyrryx: If we do this, it might even be possible to make `SingleScope` and `DecoderScope` implement `Copy` as well!

src/decoder.rs
mikeday: I think this let-else introduces a _new_ v, it doesn't assign to the existing variable.

src/decoder.rs
mikeday: I've renamed it to MultiScope to parallel SingleScope.

src/decoder.rs
mikeday: I've changed it to `&'a str` although not made it Copy as Scope itself can't be Copy.

src/decoder.rs
mikeday: I changed it to a match and it looks neat enough.

mikeday:
I think this is ready to go now!

src/decoder.rs
```diff
@@ -545,117 +544,145 @@ pub enum ScopeEntry {
...
-        let names = Vec::new();
-        let entries = Vec::new();
-        Scope {
-            parent,
-            names,
-            entries,
+    pub fn get_bindings(&self, bindings: &mut Vec<(Label, ScopeEntry)>) {
+        match self {
+            Scope::Empty => {}
+            Scope::Multi(normal) => normal.get_bindings(bindings),
```
archaephyrryx, line 593: Though purely cosmetic, the hold-over binding identifier `normal` is a bit misplaced here...

src/decoder.rs
```diff
@@ -1356,16 +1384,16 @@ mod tests {
 
     fn accepts(d: &Decoder, input: &[u8], tail: &[u8], expect: Value) {
         let program = Program::new();
-        let mut scope = Scope::new();
-        let (val, remain) = d.parse(&program, &mut scope, ReadCtxt::new(input)).unwrap();
+        let scope = Scope::Empty;
```
archaephyrryx, line 1387: This should be simple enough to inline into its single site of use, though admittedly not a huge need to do so.

src/decoder.rs
```diff
@@ -1356,16 +1384,16 @@ mod tests {
...
+        let (val, remain) = d.parse(&program, &scope, ReadCtxt::new(input)).unwrap();
         assert_eq!(val, expect);
         assert_eq!(remain.remaining(), tail);
     }
 
     fn rejects(d: &Decoder, input: &[u8]) {
         let program = Program::new();
-        let mut scope = Scope::new();
-        assert!(d.parse(&program, &mut scope, ReadCtxt::new(input)).is_err());
+        let scope = Scope::Empty;
```
archaephyrryx, line 1395: Likewise, we don't necessarily need a name for `Scope::Empty` if it is immutable and only used in one place. Unless I am wrong about lifetime inference for temporary values, that is.

src/decoder.rs
```diff
@@ -545,117 +544,145 @@ pub enum ScopeEntry {
     Decoder(Decoder),
 }
 
-pub struct Scope<'a> {
-    parent: Option<&'a Scope<'a>>,
-    names: Vec<Label>,
-    entries: Vec<ScopeEntry>,
+pub enum Scope<'a> {
```
archaephyrryx, line 547: It might be more ergonomic if we had

```rust
impl<'a> From<&'a MultiScope<'a>> for Scope<'a> {}
impl<'a> From<SingleScope<'a>> for Scope<'a> {}
impl<'a> From<DecoderScope<'a>> for Scope<'a> {}
```

which would allow us to unify on an `.into()` pattern for constructing the wrapper instead of explicitly naming the variant in question. Not sure how useful that might be, but it could make things easier, perhaps.

archaephyrryx (approved):
Modulo the few recommendations I am including in the review, which are mostly optional, this looks good to go to me as well.

src/decoder.rs
mikeday: Fixed.

src/decoder.rs
mikeday: Fixed.

src/decoder.rs
mikeday: Fixed.

src/decoder.rs
mikeday: I think I'll hold off on this for now as scope might require some other changes involving traits soon.

PR #130, opened 2023-11-20
Renaming, documenting of methods and items in the MatchTree* family

archaephyrryx:
Additionally, removes an unconditionally `Ok()` `Result` type from `MatchTreeLevel::merge_branch`.

EDIT:
Also adds a new document, `DESIGN.md`, for the purposes of externally documenting the `MatchTree*` family of types and the model they present.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -118,7 +118,7 @@ must have mutually disjoint branching-predicates. That is, for all well-formed `
 
 However, the same does not hold for `MatchTreeLevels`, as they represent the superposition of all choice-points at a common depth. If that depth is non-zero, the `MatchTree` may still be well-formed even if two choice-points within the `MatchTreeLevel` have some non-null intersection among the cartesian product of their respective lists of branching-predicates.
 
-For this reason, a `MatchTreeLevel` cannot represent a single, unambiguous descent into a follow-set for any given byte of input, but rather, must form a disjunction over the smallest set of mutually disjoint subsets of all the branching predicates of its constituent choice-points, such as in the following algorithm (in pseud-ocode):
+For this reason, a `MatchTreeLevel` cannot represent a single, unambiguous descent into a follow-set for any given byte of input, but rather, must form a disjunction over the smallest set of mutually disjoint subsets of all the branching predicates of its constituent choice-points, such as in the following algorithm (in pseudo-code):
 
 ```none
 UnifySteps(Steps: [(FormatIx, MatchTreeStep)]):
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1,0 +1,155 @@
+# MatchTree Design Documentation
+
+This document is to serve as an external explanation of the design, model, and API of the
+`MatchTree` infrastructure within `doodle`. It should be viewed as a living document, and may not
+precisely reflect the most up-to-date implementation as details change. However, it is easier to include
+longer digressions on the model of a given type or operation within the system, in a separate document,
+rather than crowd the definition sites in the library source file itself.
+
+## Terminology
+
+Abstractly, we may refer to the model of a `MatchTree` in terms of a trie whose nodes are *choice points* and whose children are *conditioned* on a particular set of values for the immediate byte of input at the current lookahead-offset into the buffer being processed.
+
+By 'choice point', we mean an individual multi-way decision between one of K possible one-step descents.
+Each such descent is associated (and conditioned on) a particular set of byte-values, which are mutually disjoint across any two branches of the same choice-point.
+
+We may refer to one or more of the byte-sets upon which a choice-point is conditioned, as the *branching-predicates* of said node.
+
+## Core Types
+
+The `MatchTree` design consists of three core types, and one auxiliary type:
+
+* [`MatchTree`](#matchtree-type-definition): a finalized prefix-tree evaluated to an abstract, but consistent depth across all branches (modulo short-circuit pruning in cases where acceptance or rejection is unconditional)
+* `MatchTreeLevel`: an aggregation of choice-points at a common descent-depth into a `MatchTree`; alternatively, the unified logic for matching a byte at a given lookahead distance into the current buffer, which may be accumulated for each depth from 0 up to the maximal lookahead and thereby form a full `MatchTree`.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1327,37 +1350,49 @@ impl<'a> MatchTreeStep<'a> {
...
+    ///
+    /// Will return `Ok(())` if said index was already marked as such, or no index was.
+    ///
+    /// Returns `Err(())` if any other index was already marked as accepting.
+    ///
+    /// # Safety
+    ///
+    /// There is no way to tell whether `index` is properly in-range within this method.
+    /// The caller must therefore either ensure it does not pass in an improper index, or
+    /// sanitizes the index returned from methods that access it further down the line.
```
mikeday, line 1395: This comment seems unnecessary.

src/lib.rs
```diff
@@ -1029,6 +1029,15 @@ struct MatchTreeLevel<'a> {
...
 
+/// A bundle of follow-sets with an externally significant index, e.g. into an array of decoders
+///
+/// While the choice-points along any given descent of a [`MatchTree`] are fully informed
+/// by sets of mutually disjoint [`ByteSet`]s, the superposition of choice-points across
+/// all descents of a fixed depth (i.e. a [`MatchTreeLevel`]) will almost invariably encompass
+/// multiple original choice-points with non-disjoint [`ByteSet`]s. In such cases, it may be necessary
+/// to create multiple virtual branches that represent a deterministic superposition in the otherwise
+/// nondeterministic single-byte-descent at that level. In such cases, we cross-reference an original index
+/// across all entries stemming from the same [MatchTreeStep].
```
mikeday, line 1040: I'm not sure I can understand this comment, and it might be better suited to external documentation.

DESIGN.md
```diff
@@ -0,0 +1,155 @@
...
+By 'choice point', we mean an individual multi-way decision between one of K possible one-step descents.
+Each such descent is associated (and conditioned on) a particular set of byte-values, which are mutually disjoint across any two branches of the same choice-point.
+
+We may refer to one or more of the byte-sets upon which a choice-point is conditioned, as the *branching-predicates* of said node.
+
+## Core Types
+
+The `MatchTree` design consists of three core types, and one auxiliary type:
+
+* [`MatchTree`](#matchtree-type-definition): a finalized prefix-tree evaluated to an abstract, but consistent depth across all branches (modulo short-circuit pruning in cases where acceptance or rejection is unconditional)
```
mikeday, line 22: What exactly makes a depth "abstract but consistent"?

DESIGN.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+To understand the significance of this type, we make the following observation:
+
+In order for a `MatchTree` to be well-formed, it and all of its descendants
+must have mutually disjoint branching-predicates. That is, for all well-formed `tree : MatchTree`, it must be the case that for all values of `b : u8`, the iterator generated by `tree.branches.filter(|(s, _)| s.contains(b)))` must yield zero or one elements only.
+
+However, the same does not hold for `MatchTreeLevels`, as they represent the superposition of all choice-points at a common depth. If that depth is non-zero, the `MatchTree` may still be well-formed even if two choice-points within the `MatchTreeLevel` have some non-null intersection among the cartesian product of their respective lists of branching-predicates.
+
+For this reason, a `MatchTreeLevel` cannot represent a single, unambiguous descent into a follow-set for any given byte of input, but rather, must form a disjunction over the smallest set of mutually disjoint subsets of all the branching predicates of its constituent choice-points, such as in the following algorithm (in pseud-ocode):
```
mikeday, line 121: Typo, pseudo-code.

DESIGN.md
```diff
@@ -0,0 +1,155 @@
...
+                        Remove all elements of Common from BS, mutably in-place
+                    Else
+                        Do nothing
+                If BS ≠ ∅
+                Then
+                    Push (BS, [(Ix, Next)]) to the end of X.branches
+                Else
+                    Do nothing
+                Append all elements of NewBranches to the end of X.branches
+    Return Some(X)
```
mikeday, line 152: I'm not sure if this pseudo-code is going to be clearer than just pointing to the Rust code in the module itself but I guess it provides an alternative view!

src/lib.rs
```diff
@@ -1418,18 +1433,25 @@ impl<'a> MatchTreeLevel<'a> {
         })
     }
 
+    /// Attempts to accumulate a `MatchTree` recursively up to an overall depth of `depth` layers,
+    /// with the immediate layer constructed based on a bundle of indexed choice-points ([`LevelBranch`]).
+    ///
+    /// If the depth limit has been reached without a decisive choice of which index to accept, returns None.
+    ///
+    /// Otherwise, returns a `Some`-wrapped `MatchTree` that is guaranteed to decide on a unique branch for
```
mikeday, line 1441: Minor point but "returns a MatchTree" is probably sufficient, with "Some-wrapped" implied.

mikeday (approved):
Some minor comments but looks good!

DESIGN.md
archaephyrryx: Perhaps there is a better way of phrasing it, but what I mean is, every branch $B_i$ is evaluated to a depth $N_i$, where $\exists N^\star \ni \forall i. N = N_i $ but the actual value of $N$ is not available, and may be unknowable in some cases (e.g. if all branches accept within $M < N$ bytes of input). If you can come up with a more direct way of getting that idea across, or if you think it is not necessary to mention, I would be happy to make the appropriate changes.

DESIGN.md
archaephyrryx: Ah, indeed. I left out the hyphen initially and clearly didn't put it in the right place :sweat_smile:

DESIGN.md
archaephyrryx: I feel that the intent of the code is easier to capture in pseudo-code, but I agree with you morally. It just might be easier to get around details of potentially opaque APIs like ByteSet. There is also no single method in the Rust code that this represents, since it is a left-fold of `merge_step` with the calls to `merge_accept` and `merge_branch` inlined. But indeed, showing the Rust code might be about as effective as writing this out.

src/lib.rs
archaephyrryx: I see your point. Will simplify this.

PR #135, opened 2023-12-01
Codegen fixes

archaephyrryx:
Introduces `ParseCtxt`, simple abstraction for threading an incrementally-consumed `ReadCtxt` through Decoders without forcing them to change return-value. This does not allow for backtracking or splitting, at least as implemented currently.

Fixes broken label sanitization rules, and suppresses them entirely for the contents of string literals.

Adds a manual test at the end of `sample_codegen.rs` to confirm that parsing can actually work.


archaephyrryx:
These changes were merged but it wasn't obvious that it happened.

PR #145, opened 2024-01-29
Change Format::UnionNondet to be consistent with Union.

mikeday:
This change shifts the variant label syntax sugar to a helper function, which simplifies the core format language.


archaephyrryx (approved):
Looks good to me.

PR #147, opened 2024-02-05
Expr::SubSeq indices must be U32.

mikeday:


archaephyrryx (approved):

src/bin/doodle/format/deflate.rs
```diff
@@ -621,10 +621,10 @@ pub fn main(module: &mut FormatModule, base: &BaseModule) -> FormatRef {
                                     Expr::Tuple(vec![
                                         tuple_proj(var("x"), 0),
                                         dup(
-                                            add(
+                                            as_u32(add(
                                                 record_proj(tuple_proj(var("x"), 1), "extra"),
                                                 Expr::U8(11),
```
archaephyrryx, line 626: Would it not be better to do `add(as_u32(record_proj(...), Expr::U32(11)))`, which avoids potential overflow? (similarly for other sites).

mikeday: I think we're safe here as the "extra" value is at most 7-bits, but ideally we could detect overflow at compile time and promote accordingly! 😆

PR #149, opened 2024-02-08
TypeChecker Engine for incremental type inference and unification of Format, Expr, Pattern

archaephyrryx:
Rebased version of Draft PR #144 

No integration into downstream (Rust codegen) or upstream (the rest of doodle) yet.


src/lib.rs
```diff
@@ -524,7 +550,7 @@ pub enum DynFormat {
 #[serde(tag = "tag", content = "data")]
 pub enum Format {
     /// Reference to a top-level item
-    ItemVar(usize, Vec<Expr>),
+    ItemVar(usize, Vec<Expr>), // FIXME - do the exprs here need type(+) info?
```
mikeday, line 553: Presumably the author should not need to specify the types of expr here, it would make more sense to specify the type of formats when declaring them if necessary.

src/lib.rs
```diff
@@ -172,39 +156,45 @@ impl ValueType {
...
             (ValueType::Record(fs1), ValueType::Record(fs2)) => {
                 if fs1.len() != fs2.len() {
-                    return Err(format!(
-                        "records must have same number of fields {fs1:?} vs. {fs2:?}"
-                    ));
+                    // field count mismatch
+                    return Err(UnificationError::Unsatisfiable(self.clone(), other.clone()));
                 }
-                // FIXME field order
+                // NOTE - because fields are parsed in declared order, two records with conflicting field orders are not operationally equivalent
```
mikeday, line 162: Is this still a problem now that record fields are sorted in the type?

mikeday: Oh sorry that's union variants, not record fields.

src/bin/doodle/main.rs
```diff
@@ -49,6 +51,8 @@ enum Command {
         /// The binary file to decode
         filename: PathBuf,
     },
+    /// Typecheck the main FormatModle
```
mikeday, line 54: s/Modle/Module/

src/lib.rs
```diff
@@ -809,10 +872,20 @@ impl FormatModule {
...
+    pub fn iter_formats(&self) -> impl Iterator<Item = (usize, Format)> + '_ {
+        (0..self.formats.len()).filter_map(|ix| {
+            let mut x_args = Vec::with_capacity(self.args[ix].len());
+            for (_, vt) in self.args[ix].iter() {
+                x_args.push(mk_value_expr(vt)?);
+            }
+            Some((ix, Format::ItemVar(ix, x_args)))
+        })
+    }
+
```
mikeday, line 888: What is this for exactly? I'm curious about the use of `mk_value_expr`.

archaephyrryx: `mk_value_expr(vt)` is a test-helper function that maps a `ValueType` to an `Expr` with that `ValueType`. It is used to feed 1 or more dummy arguments to non-nullary ItemVars, in such a way as to not trigger complaints of incorrect Expr types before the Format itself can be type-inferred.

src/bin/doodle/main.rs
archaephyrryx: Oops, fixed this just now.

mikeday (approved):

PR #151, opened 2024-02-12
Move base module into library crate

mikeday:
This allows convenience functions to be used in tests.


mikeday:
Closing this as I think the base module should be split rather than moved.

PR #153, opened 2024-02-15
Split base module into helper so functions can be used in tests

mikeday:


archaephyrryx (approved):
Looks good, definitely solves a current issue with writing format-based tests.

PR #156, opened 2024-03-20
Type-Aware Codegen with elaborated Formats

archaephyrryx:
Work-in-progress effort to fix missing type-information in generated code through TC-engine-based elaboration, using Typed* versions of Format, Expr, Pattern, Decoder, and other related types.


mikeday (approved):
The complexity is building up but let's run with it and see how it goes!

PR #157, opened 2024-03-27
Parsing engine

archaephyrryx:
Working version of an integrated Parsing model to act as a backend for generated code.

Fixes several bugs in the codegen generation module.

Adds manually-implemented test-module for generated code.


src/parser/offset.rs
```diff
@@ -0,0 +1,398 @@
...
+            // yield None instead of Some(Equal) if same bit-level offset but different modes
+            Ordering::Equal if self.is_bit_mode() ^ other.is_bit_mode() => None,
+            _ => Some(partial),
+        }
+    }
+}
+
+pub(crate) struct BufferOffset {
+    current_offset: ByteOffset, // the 'true' value of the offset, reinterpreted according to the other fields
+    view_stack: ViewStack,
```
mikeday, line 164: Would be good to know what this means, like reinterpreted according to view_stack?

mikeday (approved):
Looks good to me!

src/parser/offset.rs
archaephyrryx: This may be an artifact of an earlier version. I frankly don't know *exactly* what it is referring to, and it is probably stale.

PR #158, opened 2024-04-02
Add lookahead_bounds method

archaephyrryx:
Adds `Format` and `TypedFormat` methods `lookahead_bounds` that mirrors `match_bounds`, but returns the expected limits on bytes read, regardless of how many are consumed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -701,6 +701,53 @@ impl Format {
         }
     }
 
+    /// Conservative bounds for number of bytes read while attempting to match a format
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -701,6 +701,53 @@ impl Format {
         }
     }
 
+    /// Conservative bounds for number of bytes read while attempting to match a format
```
mikeday, line 704: Maybe we could say "conservative over-estimate" to make it clear which direction it's being conservative in.

mikeday (approved):
Looks good to me!

PR #159, opened 2024-04-07
Simplify deflate format

mikeday:
Use parameterised formats instead of Rust functions, greatly shrinking the generated code.


archaephyrryx (approved):
Definitely an improvement as far as compact generated code!

PR #160, opened 2024-04-08
[FIX] Add ItemVar arguments to typechecker eval scope of called format

archaephyrryx:


archaephyrryx:
Closed in favor of #161

PR #164, opened 2024-04-22
Replace Expr::Inflate with two new primitives

mikeday:
Expr::FlatMapList and Expr::SubSeqInflate can be combined to replace Expr::Inflate, allowing all the record matching to be done in a normal expression without special compiler support.


archaephyrryx:
This can be closed, I assume, since it is a no-op after the #172 merge...

PR #167, opened 2024-05-16
RepeatBetween format for bounded repetitions

archaephyrryx:
Adds basic, experimental support at all appropriate layers for a new primitive, `Format::RepeatBetween(N, M, F)`.

Adds Peano format for expressing numeric sequences as `(S*Z)+` strings, with test file with the first 6 digits of pi.


src/lib.rs
```diff
@@ -729,6 +735,9 @@ impl Format {
             Format::Repeat(_) => Bounds::new(0, None),
             Format::Repeat1(f) => f.lookahead_bounds(module) * Bounds::new(1, None),
             Format::RepeatCount(expr, f) => f.lookahead_bounds(module) * expr.bounds(),
+            Format::RepeatBetween(xmin, xmax, f) => {
+                f.lookahead_bounds(module) * Bounds::union(xmin.bounds(), xmax.bounds())
+            }
```
mikeday, line 740: `lookahead_bounds` might be a bit weird for `RepeatCount` and `RepeatBetween` maybe? I need to think about that one 🤔

src/lib.rs
```diff
@@ -1275,6 +1288,49 @@ impl<'a> MatchTreeStep<'a> {
         }
     }
 
+    /// Constructs a [MatchTreeStep] that accepts a repetition whose count is bounded above and below, with a trailing sequence of partially-consuemd formats ([`Next`]s)
+    ///
+    /// The format in quiestion will  an arbitrary number of times between `min` and `max`, where `minmax ::= (min, max)`
```
mikeday, line 1293: s/quiestion/question/

mikeday: Oh and s/will  an/will match an/

src/lib.rs
```diff
@@ -1479,6 +1565,59 @@ impl<'a> MatchTreeStep<'a> {
...
+                            } else {
+                                Self::from_next(module, next)
+                            }
+                        }
+                        Ordering::Greater => {
+                            panic!("incoherent repeat-between: min {} > max {}", min, max)
+                        }
+                    },
+                    _ => {
+                        unreachable!("inexact repeat-between bounds (not technically a problem but not what the combinator was designed for...");
```
mikeday, line 1617: Possibly better to allow variable/dependent bounds?

src/lib.rs
```diff
@@ -1595,6 +1734,24 @@ impl<'a> MatchTreeStep<'a> {
...
+                        Ordering::Less => {
+                            Self::from_repeat_between(module, (min, max), a, next.clone())
+                        }
+                        Ordering::Equal => Self::from_repeat_count(module, min, a, next.clone()),
+                        Ordering::Greater => {
+                            panic!("incoherent repeat-between: min {} > max {}", min, max)
+                        }
+                    },
+                    _ => {
+                        unreachable!("inexact repeat-between bounds (not technically a problem but not what the combinator was designed for...");
```
mikeday, line 1751: Similarly here.

src/bounds.rs
```diff
@@ -35,6 +35,19 @@ impl Bounds {
         }
     }
 
+    /// Dual method to `union`, which keeps the most restrictive upper-bound of the
+    /// two Bounds values rather than the least.
+    pub fn intersection(lhs: Bounds, rhs: Bounds) -> Bounds {
+        Bounds {
+            min: usize::min(lhs.min, rhs.min),
```
mikeday, line 42: This seems a little misleading, wouldn't a true intersection take the max of the min values, as well as the min of the max values? Perhaps this needs a more specific name (intersect_upper or intersection_max or something). Where specifically is it used?

mikeday (approved):
Looks good! The new `RepeatBetween` primitive should be helpful and the way it's implemented with `RepeatMax` is clear to read. Some minor review comments attached.

src/lib.rs
archaephyrryx: Right...let me know if you happen to figure out what it should be (assuming that doesn't happen to be what it is at the moment)...

src/lib.rs
archaephyrryx: Same as above, it is hard to figure out without an Expr-evaluator what number of repeats are 'acceptable' in a dynamic context. Say, for example:

```rust
record([
    ("min", base.u32be()),
    ("delta", base.u32be()),
    ("seq", repeat_between(var("min"), Expr::Arith(Arith::Add, Box::new(var("min")), Box::new(var("delta"))), base.u8())
])
```

How would one construct a match-tree for this?

src/bounds.rs
archaephyrryx: This is a bit of an artifact, from `LengthLimited` which ended up getting axed early. The intention was to say:

$$\mathrm{MatchBounds}(\verb!LengthLimited!(\mathit{X}, \mathit{F}) \coloneqq \verb!Bounds::intersection!(\verb!Expr::bounds!(\mathit{X}), \verb!Format::match_bounds!(\mathit{F}))$$

But that never got surfaced as LengthLImited was a bit complicated to implement properly and wasn't ultimately necessary for the PNG case we were working on because RepeatBetween was enough,

src/lib.rs
mikeday: I guess it's only a conservative over-estimate, so it's fine. But if you had `RepeatCount(expr, f)` then while `match_bounds` is `f.match_bounds(module) * expr.bounds()`, `lookahead_bounds` could be tighter than `f.lookahead_bounds(module) * expr.bounds()`, it could be something like `f.match_bounds() * (expr - 1) + f.lookahead_bounds()`, where we add up all the bytes we actually read to match the first N-1 repetitions and then the bytes we looked ahead for the last one (which will be >= the number of bytes we matched). However this is a completely unnecessary thing to do, so let's not do it.

src/lib.rs
mikeday: I think we just shrug and cut off the match tree, much as we do for e.g. `RepeatUntilLast` and `RepeatUntilSeq`; just means you can't switch on something that starts with a `RepeatBetween` of unknown size, which is fine, it's still a legal format to have in cases where you don't need to switch on it.

src/bounds.rs
mikeday: Right, can always rename it or comment it out as appropriate.

src/lib.rs
archaephyrryx: Fixed these on a different PR (#168).

src/lib.rs
archaephyrryx: I left a comment on (#168) to revisit these lines, hopefully nothing should break in the meantime so there oughtn't be a pressing need to change this right away. If necessary, we can always return to this. Will mentally bookmark this as well.

PR #168, opened 2024-05-22
Static analysis for selective elision of vacuous expressions, promotion of hard-coded strings to productive elements of associated model-structures

archaephyrryx:
Gets rid of the `#![allow(dead_code)]` attribute at the root of `rust_ast.rs` to surface unused variants and methods (and removes them).

Fixes many proper typos in comments, and changes identifiers in certain places to keep spellchecking simple.

Basic heuristics to selectively elide vacuous expressions like

```rust
let _ = ();
```

from generated code by determining, using a loose heuristic, whether any given `RustExpr` is 'pure' in the Haskell sense. 

Promotes any hard-coded strings leftover in `rust_ast.rs` or `codegen/mod.rs` that inject 'extra' content into standard productions, into associated fields of the types being augmented, which are then expanded as first-class elements of the Rust-AST model rather than hacky afterthoughts.

### DIFF
--- a/codegen_tests.rs
+++ b/codegen_tests.rs
@@ -4,14 +4,14 @@ use super::*;
 
 type TestResult<T = ()> = Result<T, Box<dyn Send + Sync + std::error::Error>>;
 
-// Stablization aliases to avoid hard-coding shifting numbers as formats are enriched with more possibilities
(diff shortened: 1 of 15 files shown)


mikeday (approved):
That looks fine.

PR #173, opened 2024-06-18
Draft: Benchmarking and profiling for slow-running decoders (e.g. Inflate)

archaephyrryx:


archaephyrryx:
The commits for this PR ended up getting merged in #174 so we don't need to keep it open.

PR #174, opened 2024-06-20
Optimizes `Expr::eval` pessimal performance by removing all ParseLoc-related code 

archaephyrryx:
Inherits history from #173


src/decoder.rs
```diff
@@ -1018,9 +512,7 @@ impl Expr {
                     for v in values {
                         let arg = Value::Tuple(vec![Value::Seq(vs), v]);
                         // TODO can we avoid cloning arg here?
```
mikeday, line 522: Can remove this comment now that arg is not cloned 🤓

mikeday (approved):
Looks good!

src/decoder.rs
archaephyrryx: Indeed, fixing it now.

PR #177, opened 2024-07-01
Format::Where to apply boolean conditions to values.

mikeday:
This adds a format for boolean assertions. It would help if we had expressions for boolean or/and/not.

There is also a new module for checking if two expressions are disjoint, intended to be used to disambiguate unions that have disjoint assertions on their branches, however it is not used yet.


archaephyrryx:
The commits in this PR were included in the recently-merged #180 so we can close this.

PR #179, opened 2024-07-01
Pattern::Int(Bounds) to simplify matching a range of integers.

mikeday:
Not used or tested anywhere yet, could be used in deflate format. Codegen is unimplemented and type checking could use a look.


archaephyrryx:
The commits in this PR were included in the recently-merged #181 so we can close this.

PR #180, opened 2024-07-02
Extend codegen support to Format::Where

archaephyrryx:
Downstream changes to #177 including codegen support and application of new helper `where_lambda` to PNG iCCP validation (rather than hardcoding `:= 0` at the syntax-level)


mikeday (approved):
That looks fine.

PR #181, opened 2024-07-02
Pattern bounds plus codegen

archaephyrryx:
Haven't yet tested that the codegen productions for `Pattern::Int` work correctly, but everything is stable as-is.


mikeday (approved):
Looks good!

PR #189, opened 2024-07-15
Improved implementation of BitAnd/BitOr traits on Bounds type

archaephyrryx:


mikeday (approved):
Looks good to me! Assuming it passes tests 🤓

PR #190, opened 2024-07-16
Renames interpreter-based `ParseError`/`ParseResult` to avoid shadowing other definition using same name

archaephyrryx:
We currently have two types called `ParseError`:
 - [doodle::error::ParseError](https://github.com/yeslogic/doodle/blob/6d46d76af3e60387a8dec5321585920f342a31f1/src/error.rs#L11)
 - [doodle::parser::error::ParseError](https://github.com/yeslogic/doodle/blob/6d46d76af3e60387a8dec5321585920f342a31f1/src/parser/error.rs#L5)

This PR changes the naming conventions of the former to use `Decode` as the action-verb instead of `Parse`, in all sites that define the type itself or its derived/variant relatives.

### DIFF
--- a/doodle-formats/src/format.rs
+++ b/doodle-formats/src/format.rs
@@ -59,10 +59,10 @@ mod test {
     use std::borrow::Cow;
 
     use super::*;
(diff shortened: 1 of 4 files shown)


mikeday (approved):
That looks fine.

PR #198, opened 2024-07-30
Implements ForEach, applies to JPEG dht-data

archaephyrryx:


generated/sample_codegen.rs
```diff
@@ -17794,22 +17794,14 @@ fn Decoder146<'input>(
...
             let mut accum = Vec::new();
-            while _input.remaining() > 0 {
-                let matching_ix = {
-                    _input.open_peek_context();
-                    _input.read_byte()?;
-                    {
-                        let ret = 0;
-                        _input.close_peek_context()?;
-                        ret
+            for n in num_codes.clone() {
```
archaephyrryx, line 17797: This clone is theoretically necessary since we can't tell in advance whether the Expr-sequence we are iterating over is Copy or can be used referentially, but in this particular case, it might be avoidable in this particular case with specific care...

doodle-formats/src/format/jpeg.rs
```diff
@@ -75,7 +75,10 @@ pub fn main(module: &mut FormatModule, base: &BaseModule, tiff: &FormatRef) -> F
         record([
             ("class-table-id", class_table_id.clone()),
             ("num-codes", repeat_count(Expr::U8(16), base.u8())),
-            ("values", repeat(base.u8())), // TODO - List.map num-codes (\n => repeat-count n u8);
+            (
+                "values",
+                for_each(var("num-codes"), "n", repeat_count(var("n"), base.u8())),
+            ), // TODO - List.map num-codes (\n => repeat-count n u8);
```
mikeday, line 81: That looks good! Is the TODO still necessary?

src/decoder.rs
```diff
@@ -855,6 +856,11 @@ impl<'a> Compiler<'a> {
                 let da = Box::new(self.compile_format(a, next)?);
                 Ok(Decoder::RepeatUntilLast(expr.clone(), da))
             }
+            Format::ForEach(expr, lbl, a) => {
+                // FIXME probably not right
```
mikeday, line 860: Perhaps this comment (and the others like it) should clarify that it's the `next` value we're passing in that is probably not right 🤔

src/lib.rs
```diff
@@ -776,6 +778,8 @@ impl Format {
                 .unwrap(),
             Format::Dynamic(_name, _dynformat, f) => f.match_bounds(module),
             Format::Apply(_) => Bounds::at_least(1),
+            // FIXME - do we have any way of approximating this better?
+            Format::ForEach(_expr, _lbl, _f) => Bounds::any(),
```
mikeday, line 782: expr bounds * f bounds?

src/lib.rs
```diff
@@ -870,6 +876,8 @@ impl Format {
             Format::Match(_, branches) => branches.iter().any(|(_, f)| f.depends_on_next(module)),
             Format::Dynamic(_name, _dynformat, f) => f.depends_on_next(module),
             Format::Apply(..) => false,
+            // REVIEW - is this correct?
+            Format::ForEach(_expr, _lbl, f) => f.depends_on_next(module),
```
mikeday, line 880: I think this is correct yes, if we permit f to be flexible, which we probably do.

src/lib.rs
```diff
@@ -1631,6 +1650,15 @@ impl<'a> MatchTreeStep<'a> {
                     Rc::new(Next::Repeat(MaybeTyped::Typed(a), next.clone())),
                 ))
             }
+            TypedFormat::ForEach(_, _expr, _lbl, a) => {
+                // FIXME - this may not be right
```
mikeday, line 1654: I think this is a reasonable approximation.

src/lib.rs
```diff
@@ -1831,6 +1859,15 @@ impl<'a> MatchTreeStep<'a> {
                     Rc::new(Next::Repeat(MaybeTyped::Untyped(a), next.clone())),
                 ))
             }
+            // FIXME - this may not be exactly correct
```
mikeday, line 1862: Similarly this comment isn't really adding much.

mikeday (approved):
Looks good!

src/decoder.rs
archaephyrryx: Indeed. I think this one might even be correct enough we don't need the comment.

src/lib.rs
mikeday: Oh right yes, we would need bounds on sequences and we don't have that right now and won't for a while. The comment can probably say "can't do better without sequence length bounds" or similar.

PR #206, opened 2024-08-23
Fix undiscovered endianness bug in deflate fixed distance-codes

archaephyrryx:


doodle-formats/src/format/deflate.rs
```diff
@@ -30,6 +30,17 @@ fn bits_value_u8(name: &'static str, n: usize) -> Expr {
...
+    if n > 1 {
+        bit_or(
+            tuple_proj(var(name), n - 1),
+            shl_u8(dist_value_u8(name, n - 1), 1),
+        )
+    } else {
+        tuple_proj(var(name), 0)
+    }
+}
+
```
mikeday, line 43: Maybe we could rename this and the existing bits function to indicate their endianness explicitly?

doodle-formats/src/format/deflate.rs
```diff
@@ -1204,17 +1224,26 @@ pub fn main(module: &mut FormatModule, base: &BaseModule) -> FormatRef {
...
+                        ),
+                        (
+                            Pattern::U8(1),
+                            fmt_variant("fixed_huffman", fixed_huffman.call()),
+                        ),
+                        (
+                            Pattern::U8(2),
+                            fmt_variant("dynamic_huffman", dynamic_huffman.call()),
+                        ),
+                        // (Pattern::U8(3), Format::Fail), // Reserved - never constructed due to Fail, so no variant needed
```
mikeday, line 1243: Is this a semantic change?

mikeday (approved):
Good work!

doodle-formats/src/format/deflate.rs
archaephyrryx: It's a bit complicated to specify the endianness, since the bits are written from LSB-to-MSB within each byte during the bitstream-to-bytestream conversion. The endianness of distance-codes is then 'natural-order' in the order a prefix code would normally be written, and proper 'numbers' have their bits written in order of LSB-to-MSB.

So one could argue that distance-codes are big-endian in a little-endian framing, or little-endian in a big-endian framing.
And equally as well, that numbers are little-endian in a little-endian frame, or big-endian in a big-endian frame.

So naming them closer to the 'bit-order' semantics like the dichotomies of  'code' vs 'number' or 'natural' vs 'reversed' would perhaps be clearer? Idk.

archaephyrryx: I do agree clearer names would be useful, but the complexity of the problem-space makes it hard to come up with unambiguous labels for the two modes of bit-writing.

doodle-formats/src/format/deflate.rs
archaephyrryx: There is no difference in the semantics aside from the commented-out line 1243, which would only be possible to write in the revised syntax since it doesn't need a variant and probably shouldn't get one (as Fail has no associated value to wrap in a Variant). If 1243 were uncommented it would be a semantic change, in explicitly marking type `0b11` as a failure-case rather than simply just an unexpected branch of the match-tree. Which probably wouldn't be a meaningful change aside from the error message that is returned, which are often discarded in favor of the error message returned by the final format in the NondetUnion covering the top-level alternation over each file-format.

archaephyrryx: We might, at some point, want to distinguish strong-local errors from generic "oops this doesn't look right, try something else". Since we might, at a certain depth into a speculative parse as a given high-level Format, have high confidence that 'this is where we should be' and that specific errors should be treated as 'correct format choice, bad data' rather than 'invalidated speculative parse', which is the only real semantic class of errors we currently have a concept of.

mikeday: Oh yes I see, no need to wrap fail in a variant, that makes sense.

PR #207, opened 2024-08-26
Expr::SeqIndex for array/vector projections

archaephyrryx:


src/helper.rs
```diff
@@ -644,3 +645,114 @@ pub fn concat(xs: Expr) -> Expr {
...
+        ],
+    )
+}
+
+/// Performs a fallible destructuring of the provided `expr` within the Format layer,
+/// either extracting the contents of a `Some(_)` value, or poisoning the current parse-state
+/// via `Format::Fail` if it is `None`.
+pub fn fmt_unwrap(expr: Expr) -> Format {
+    Format::Match(expr, vec![(pat_some(bind("x")), Format::Compute(var("x"))), (pat_none(), Format::Fail)])
+}
```
archaephyrryx, line 723: While this is relatively easy to erase using `Format::LetFormat` or its helper `chain(F0, Label, F1)`, we might want to think about whether there is a 'good way' to produce a `Fail`-like behavior within the pure Expr layer, possibly restricted to `Compute(X)`.

archaephyrryx: E.g.

```rust
pub enum Expr {
    /* everything we already have */
   Undefined,
}
```

where

$$
\mathrm{Decode}(\mathtt{Format::Compute}(\mathtt{Expr::Undefined})) \simeq \mathrm{Decode}(\mathtt{Format::Fail})
$$

archaephyrryx: This might completely gum up the model though, so while forcing us to lift the unwrap into the Format layer is a bit awkward, it might be an easier approach overall.

mikeday: Non-terminating expressions do raise questions about order of evaluation etc. but arguably so does division by zero.

src/loc_decoder.rs
```diff
@@ -816,6 +816,19 @@ impl Expr {
...
             },
+            Expr::SeqIx(seq, index) => {
+                match seq
+                    .eval_with_loc(scope)
+                    .coerce_mapped_value()
+                    .get_sequence()
+                {
+                    Some(values) => {
+                        let index = index.eval_value_with_loc(scope).unwrap_usize();
+                        Cow::Owned(values[index].clone())
```
mikeday, line 827: Can this return a reference to the value or is it not possible here?

mikeday (approved):
Looks good!

src/loc_decoder.rs
archaephyrryx: I do suppose it could, in point of fact. I will double-check that it works before I push that change.

archaephyrryx: Ah, no, it actually can't be referenced. Because the capture of `values` lives behind the stack-local evaluation of `.eval_with_loc` which may be an owned value due to the `Cow` and therefore will not live long enough to be referenced. There may be some pattern-matches we could apply to possibly obtain a longer-lived reference to introspect and return, but that is arguably more work than is really worth bothering with...

archaephyrryx: It is actually possible, but only if we do this:

```rust
            Expr::SeqIx(seq, index) => {
                let eval = seq.eval_with_loc(scope);
                match eval {
                    Cow::Borrowed(v) => match v.coerce_mapped_value().get_sequence() {
                        Some(values) => {
                            let index = index.eval_value_with_loc(scope).unwrap_usize();
                            Cow::Borrowed(&values[index])
                        }
                        _ => panic!("SeqIx: expected Seq"),
                    }
                    Cow::Owned(v) => match v.coerce_mapped_value().get_sequence() {
                        Some(values) => {
                            let index = index.eval_value_with_loc(scope).unwrap_usize();
                            Cow::Owned(values[index].clone())
                        }
                        _ => panic!("SeqIx: expected Seq"),
                    }
                }
            }
```

archaephyrryx: We could clean it up, e.g. factoring out the duplicate evaluation of `index`, and maybe even find a way of producing this sort of thing auto-magically using either a helper function or a macro (though the latter might be a bit heavy-handed for this sort of one-off thing)

mikeday: Probably not worth overthinking, just wanted to check!

archaephyrryx: I think I managed to finesse it a bit:

```rust
            Expr::SeqIx(seq, index) => {
                cow_map(seq.eval_with_loc(scope), |v| {
                    match v.coerce_mapped_value().get_sequence() {
                        Some(values) => {
                            let index = index.eval_value_with_loc(scope).unwrap_usize();
                            &values[index]
                        }
                        _ => panic!("SeqIx: expected Seq"),
                    }
                })
            }
```

with local function

```rust
fn cow_map<'a, T, U>(x: Cow<'a, T>, f: impl for<'i> Fn(&'i T) -> &'i U) -> Cow<'a, U>
where
    T: 'static + Clone,
    U: 'static + Clone + ToOwned,
{
    match x {
        Cow::Borrowed(x) => Cow::Borrowed(f(x)),
        Cow::Owned(x) => Cow::Owned(f(&x).to_owned()),
    }
}
```

mikeday: neat! almost feels like something like that should be in the standard library 🤔

archaephyrryx: Also noticed that this same helper function can be used on `RecordProj` and `TupleProj`, as well as `SeqIx` in both `decoder.rs` and `loc_decoder.rs`.

PR #210, opened 2024-08-28
Expr::LeftFold + assorted helpers

archaephyrryx:
Also cleans up residual workload from `lambda_tuple` PR #209

### DIFF
--- a/src/helper.rs
+++ b/src/helper.rs
@@ -1,6 +1,6 @@
 use crate::bounds::Bounds;
 use crate::byte_set::ByteSet;
-use crate::{Arith, Expr, Format, IntRel, IntoLabel, Label, Pattern, ValueType};
+use crate::{Arith, BaseType, Expr, Format, IntRel, IntoLabel, Label, Pattern, ValueType};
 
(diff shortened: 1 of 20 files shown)


src/helper.rs
```diff
@@ -777,11 +786,189 @@ pub fn seq_proj(seq: Expr, index: usize) -> Expr {
...
         ),
     )
 }
+
+/// Shorthand for Expr::LeftFold
+pub fn left_fold(f: Expr, init: Expr, init_vt: ValueType, seq: Expr) -> Expr {
+    Expr::LeftFold(Box::new(f), Box::new(init), init_vt, Box::new(seq))
+}
+
+/// Computes the larger of two given `Expr`s, let-biased if equal
```
mikeday, line 804: left-biased?

mikeday (approved):
That looks fine, we should probably avoid adding too many primitives we don't actually use though.

src/helper.rs
archaephyrryx: Oops! Nice catch :+1:

PR #219, opened 2024-09-18
Split out ValueType, Rc-based TypeHint for Expr, Box Expr in Format

archaephyrryx:
Significantly reduces the stack footprint of owned values of type Format, Expr by using smart-pointer indirection for their largest arguments (ValueType for Expr, Expr for Format)

### DIFF
--- a/doodle-formats/src/format/deflate.rs
+++ b/doodle-formats/src/format/deflate.rs
@@ -61,6 +61,7 @@ fn dist_value_u8(x: &'static str) -> Expr {
 ///
 /// Auto-recursive for n > 1, using a successively smaller prefix of the tuple expression.
 fn bits_value_u16(name: &'static str, n: usize) -> Expr {
+    // FIXME - try to balance the tree with explicit conditional flow over value of n
     if n > 1 {
         bit_or(
             shl_u16(tuple_proj(var(name), n - 1), cast(n - 1)),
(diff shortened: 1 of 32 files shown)


doodle-formats/src/format/deflate.rs
```diff
@@ -83,7 +83,7 @@ fn dist8(base: &BaseModule) -> Format {
 /// standard LSB-to-MSB write order for numeric values in the DEFLATE specification.
 fn bits8(n: usize, base: &BaseModule) -> Format {
     if n == 0 {
-        return Format::Compute(Expr::U8(0));
+        return Format::Compute(Box::new(Expr::U8(0)));
```
mikeday, line 86: I guess a helper `compute` function might not hurt.

doodle-formats/src/format/jpeg.rs
```diff
@@ -17,7 +17,7 @@ pub fn main(module: &mut FormatModule, base: &BaseModule, tiff: &FormatRef) -> F
             ("length", base.u16be()),
             (
                 "data",
-                Format::Slice(sub(var("length"), Expr::U16(2)), Box::new(data)),
+                Format::Slice(Box::new(sub(var("length"), Expr::U16(2))), Box::new(data)),
```
mikeday, line 20: Similarly a slice helper if we don't already have one.

mikeday (approved):
That looks fine.

doodle-formats/src/format/deflate.rs
archaephyrryx: Yeah, I thought about this too as I was changing all of the unboxed Expr to be boxed...will author a new commit that hopefully simplifies such common-cases.

doodle-formats/src/format/jpeg.rs
archaephyrryx: Similarly fixing this.

PR #225, opened 2024-10-04
Draft: Add type-informed decoder names in codegen layer

archaephyrryx:


mikeday (approved):
Looks like a solid improvement!

PR #232, opened 2025-01-13
Opentype ItemVariationStore implementation

archaephyrryx:
Updated so that it is no longer floating, and should be possible to directly merge onto main


mikeday (approved):
Looks good to me!

PR #238, opened 2025-02-10
Codegen Refactor: Context-Aware syntax-elision and GenLambda

archaephyrryx:

PR #239, opened 2025-03-17
Redesign syntactical conventions for specifying `Record`-kinded parses

archaephyrryx:
(Includes commits from #238)

Removes `Format::Record` entirely, and introduces two new `Format` variants to serve for record-construction purposes:
- `MonadSeq`: Variation of `LetFormat` that explicitly discards one parse and yields the result of the other (i.e. without binding to a variable)
- `Hint`: Used to associate a particular informational bread-crumb with a `Format` node, so that deep introspection and complex heuristics are not needed to identify the meta-structural properties of a `Format`. Currently used to wrap the first bind-or-sequence of a Record construction, so that it can be identified as being Record-style far more simply.

This shift introduces a slight distinction to Record-kinded parses, between **old-style** and **new-style**:

Old-style records bind (and persist, at value-level) every pseudo-field (i.e. each LHS-format of a `LetFormat`) with an associated name, and the names/order of the individual pseudo-fields dictate the names/order of the persisted record-fields.

New-style records are a super-set of old-style-records, in which any of the following are permitted (and the presence of at least one means that the record is new-style)
- Unbound pseudo-fields (i.e. any use of `MonadSeq` in the right-associative spine)
- Ephemeral pseudo-fields (those which are bound in order to be used in computations or dependent-parses later on, but which are not stored in the final Record)
- Record field re-ordering or re-naming (the names and order-of-definition of the fields of the record vary even slightly from the binding-names and nesting-order

(All old-style records are new-style records, but in some cases 'new-style' is used to refer to cases that do not fall into old-style.)

Several supporting types/methods have been added, and some existing methods modified, in light of this redesign:
- Added type: `FieldLabel`, which is used to represent the capture-level details of a pseudo-field in a new-style (including old-style) record format
- Added type: `RecordFormat`, which is used to encapsulate the list of pseudo-field (as labels and formats) of any new-style record format
 
## Breaking Changes

Due to how `Format::record` has been re-implemented, the argument passed in must satisfy `<Self as IntoIterator>::IntoIter: DoubleEndedIterator`. In theory, this constraint could be lifted by using a Cons-list implementation, VecDeque, or LinkedList internally for fields-as-LIFO purposes, but in practice most field-lists will be given as array-like (`[_; N]`, `&[_]`, `Vec<_>`, ...) and so this is not a high priority

## Usage Patterns 

The basic construction used for parsing Records under this model is lightly inspired by Haskell `do` syntax, where the record-construction itself is a pure computation (`Expr`) that uses the value-bindings produced via a sequence of  impure computations (`Format`):

### Example:

Before:
```rust
Format::Record(vec![
    ("__x", x_format), // Semi-anonymous
    ("_y", y_format), // Ephemeral
    ("z", z_format(var("_y"))), // Permanent
])
```

Stylized:
```haskell
oldStyle = do
    __x <- xFormat
    _y <- yFormat
    z <- zFormat _y
   compute $ Record { __x = __x, _y = _y, z = z }

newStyle = do
    xFormat
    y <- yFormat
    z <- zFormat y
   compute $ Record { z = z }
```

Verbatim
```rust
// old-style
chain(x_format, "__x",
    chain(y_format, "_y",
        chain(z_format(var("_y")), "z",
            compute(Expr::Record(vec![("__x", var("__x")), ("_y", var("_y")), ("z", var("z"))])))
        )
    )
)

// new-style
monad_seq(x_format,
    chain(y_format, "y",
        chain(z_format(var("y")), "z",
            compute(Expr::Record(vec![("z", var("z"))]))
        )
    )
)
```

### DIFF
--- a/doodle-formats/src/format/png.rs
+++ b/doodle-formats/src/format/png.rs
@@ -408,12 +408,11 @@ pub fn main(
                     "length",
                     expr_lte(var("length"), Expr::U32(0x7fff_ffff)),
                 ),
-            ), // NOTE: < 2^31
+            ),
             (
                 "tag",
-                chain(
+                monad_seq(
                     Format::PeekNot(Box::new(is_byte(b'I'))),
-                    "_",
                     repeat_count(Expr::U32(4), base.ascii_char_strict()),
                 ),
             ),
--- a/generated/api_helper/otf_metrics.rs
+++ b/generated/api_helper/otf_metrics.rs
@@ -1709,7 +1709,7 @@ enum DeltaValues {
 ///
 /// # Examples
 ///
-/// ```no_run
+/// ```ignore
 /// // bits::<8>(raw) is the same as `raw as i8` so we omit those cases
 ///
 /// // We only show the significant endpoints of the positive and negative ranges
--- a/generated/codegen_tests.rs
+++ b/generated/codegen_tests.rs
@@ -10,7 +10,7 @@ fn test_png_signature_decoder() {
     // PNG signature
     let input = b"\x89PNG\r\n\x1A\n";
     let mut parser = Parser::new(input);
-    let ret = Decoder103(&mut parser);
+    let ret = Decoder158(&mut parser);
     assert!(ret.is_ok());
 }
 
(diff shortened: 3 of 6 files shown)


mikeday (approved):
That looks fine.

PR #240, opened 2025-03-19
First-Class `Option::Some` and `Option::None` in `Format`

archaephyrryx:
(Includes commits from #239 )

Introduces novel `Format` variant `Format::LiftedOption :: Option<Box<Format>> -> Format`.

Re-implements `fmt_some` and `fmt_none` helpers in `src/helper.rs` to use `Format::LiftedOption` instead of `Format::Map`.


mikeday (approved):
Looks good to me!

PR #245, opened 2025-04-09
Fix `local_try` to only generate closures if stmts short-circuit

archaephyrryx:
Prevents generation of `(|| DecoderXYZZY(_input))()?` one-liners where closure is constructed and immediately called without any difference in behavior.


mikeday (approved):
Looks fine I think?

PR #253, opened 2025-05-05
Fix Clippy 'ptr-arg' lint

archaephyrryx:
Removes RLE format from exposed module definitions because it is the solitary exception to the rule that we never need `&Vec` rather than the preferred `&[_]`


mikeday (approved):
That looks fine.

PR #255, opened 2025-05-07
Demote PNG Signature format to avoid complex type in decoder signature

archaephyrryx:


mikeday (approved):
That looks okay, although sometimes it's nice to see the signature in the output.

archaephyrryx:
> That looks okay, although sometimes it's nice to see the signature in the output.

To keep things simple I left it out, but if we wanted, we could redefine `is_bytes` to return `[u8; N]` for appropriate `N`, though that would require a number of changes in the doodle library (and codegen, but more than just codegen).

PR #262, opened 2025-05-16
Add Format::Sequence, Expr::Append for Sequence QoL

archaephyrryx:
Adds `Format::Sequence`, which is effectively a restricted form of `Format::Tuple` for parses of the same type, which returns them in a `Seq` container rather than an N-tuple.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -184,10 +184,13 @@ pub enum Expr {
     Dup(Box<Expr>, Box<Expr>),
 
     LiftOption(Option<Box<Expr>>),
+    /// [T] -> [T] -> [T] (append operation)
+    Append(Box<Expr>, Box<Expr>),
 }
 
 impl Expr {
     pub const UNIT: Self = Expr::Tuple(Vec::new());
+    pub const NIL: Self = Expr::Seq(Vec::new());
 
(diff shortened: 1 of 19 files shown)


src/lib.rs
```diff
@@ -538,6 +541,16 @@ impl Expr {
...
+            Expr::Append(lhs, rhs) => {
+                let lhs_type = lhs.infer_type(scope)?;
+                let rhs_type = rhs.infer_type(scope)?;
+                if !matches!(lhs_type, ValueType::Seq(_)) {
+                    return Err(anyhow!("Append: lhs is not Seq: {lhs_type:?}"));
+                } else if !matches!(rhs_type, ValueType::Seq(_)) {
+                    return Err(anyhow!("Append: rhs is not Seq: {rhs_type:?}"));
+                }
+                Ok(ValueType::Seq(Box::new(lhs_type)))
+            }
```
mikeday, line 553: Does this need to check that the two sequences have the same element type?

archaephyrryx: In looking over this, not only are you 100% correct about the missing unification-check, the return-value is incorrectly wrapped in an extra layer of Seq.

mikeday (approved):
That looks fine to me.

PR #266, opened 2025-08-04
Implement ViewFormat::ReifyView

archaephyrryx:


src/read.rs
archaephyrryx, line 1: These changes were temporarily necessary while considering embedding `ReadCtxt` inside of `Value::View`, before deciding to just store the offset (and not a slice of the buffer) in said variant. We can probably discard these changes, though they are probably harmless.

mikeday (approved):
That looks fine.

PR #295, opened 2026-04-13
Add support for NumVar in numeric trees

archaephyrryx:
NumVar doodle->analytic embeds are plausibly working, but signed-int values can't be used flexibly in the native arithmetic model of the doodle core grammar.


mikeday (approved):
Gets difficult to review as the codebase is quite complex at this point 🤔

PR #296, opened 2026-04-21
Use signed-int constructions in Opentype format definition

archaephyrryx:
Phases out scaffolding `s8|s16be|s32be|s64be` in `opentype.rs` and uses constructed I8..I64 parsers instead.

Also modifies `otf_metrics` accordingly based on updated types.

### DIFF
--- a/doodle-formats/src/format/opentype.rs
+++ b/doodle-formats/src/format/opentype.rs
@@ -6839,7 +6839,7 @@ pub(crate) mod head {
         ]);
 
         // NOTE - Should be 2 for modern fonts but we shouldn't enforce that too strongly
-        /* ConstEnum(s16be) {
+        /* ConstEnum(i16be) {
          *     Mixed    =  0,
          *     StrongLR =  1,
          *     WeakLR   =  2,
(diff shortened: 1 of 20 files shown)


doodle-formats/src/format/opentype.rs
```diff
@@ -286,26 +284,22 @@ mod util {
         fmt_variant("F2Dot14", u16be())
     }
 
-    // FIXME[epic=signedness-hack]
-    /// Scaffolding to signal intent to use i8 format before it is implemented.
+    /// Parses a u8 value and performs a bitwise cast to i8.
     pub(crate) fn s8() -> Format {
         u8()
     }
```
mikeday: This says it performs a bitwise cast?

archaephyrryx: It's leftover from a blanket change that I ended up reverting and substituting line-by-line to make any type-inference failures easier to track down. The function itself isn't actually called anywhere anymore, and `s16be` is called in only one place (`head` table `glyphDirHint`). I can remove the unused functions and change the comment on `s16be`, though.

archaephyrryx: I went ahead and removed s16be as well, after giving slightly better support for `font_direction_hint`

mikeday (approved):
Looks good!

PR #305, opened 2026-05-14
Fixes for various technical debt

archaephyrryx:
Improved support for `Value::Numeric`

Extends `Expr::bounds` to cover more cases

Prospectively removes `Map` over `Format::Pos` in `pos32` due to changed typechecker details

Fixes bug caused by an imprecise stand-in value for `Next` when compiling `RepeatCount` and similar runtime-checked repetition formats.


mikeday (approved):
That looks fine.

PR #315, opened 2026-06-14
Opentype SVG Implementation

archaephyrryx:
Implementation of `'SVG '` table, treating SVG Documents as uninterpreted byte-arrays.


mikeday (approved):
Looks good!

PR #319, opened 2026-07-21
Extends `ViewFormat::ReadArray` to ad-hoc types

archaephyrryx:
Extends the `ViewFormat::ReadArray` to apply not only to MarkerType primitives (e.g. `U32Be`), but also to 

- FormatRef calls that expand to `CommonOp` parses
- FormatRef calls that expand to single-variant enum wrappers over `CommonOp` parses
- Records all of whose fields are either themselves `CommonOp` parses, or either of the previous two categories (and which do not contain any anonymous/temporary fields that advance the parse-offset)

```rust
let foo = module.define_format("foo", u32be());
let bar = module.define_format("bar", fmt_variant("Bar", u16le()));

record([
    ("x", u8()),
    ("y", foo.call()),
    ("z", bar.call()),
])
```

Adds sufficient machinery in the code-generation layer to model the `ReadUnchecked` trait from `allsorts` and conditionally produce implementation-blocks of `ReadUnchecked` on any ad-hoc types that are directly parametrized in a `ReadArray` context, as well as on any lower-level ad-hoc types they embed that would need similar implementations (e.g. `enum { Bar(u16) }` in the example above).

Exercises this new support by selectively re-implementing certain `repeat_count(expr, fixed_format)` patterns in `opentype.rs` and its submodules with `read_array(expr, fixed_format)` (using FormatRef or CommonOp parses as appropriate).

It is important to note that changing from `repeat_count(expr, format.call())` to `from_here(read_array(expr, format))` is **not** a guaranteed-stable change, because the former explicitly advances the parse-buffer by the length of the overall array, and the latter does not advance the parse-buffer whatsoever. Therefore, in contexts where the exact offset must be explicitly tracked (e.g. when the element being changed is a component parse in the 'spine' of a format being read via `repeat_count` or similar, itself). It is possible to get around this using `slice` with the appropriate `expr * <stride>` computation, but the `from_here` glue already adds a slight complexity overhead to what would otherwise be a simple `repeat_count` and so it may not be worthwhile to engineer a solution for constructing the correct slice.

## OpentypeModule

As part of the refactoring from `repeat_count` to `read_array` in certain places, some of the ubiquitous simple opentype-related formats (e.g. `f2dot14`) were promoted from 'constructed-everywhere' (`fn() -> Format`) to 'defined-and-called' (`fn(&mut FormatModule) -> FormatRef`). To reduce the impact of this change, a new model for binding already-defined `FormatRef` values to a single 'global context' object was introduced in parallel. The type in question is `OpentypeModule`, which is used in place of `FormatModule` (as `&mut OpentypeModule<'_>`) in various larger-scale format-declaration factory functions. This struct also contains a new dictionary-state object `OpentypeDictionary`, which contains various `OnceCell<FormatRef>` fields for the most-widely-used format-refs that would otherwise have to be passed in as separate parameters to every per-feature-table function that would need to make references to them. This dictionary includes both already-FormatRef cases like `ClassTable` and `ItemVariationStore`, as well as newly-minted FormatRef cases like `F2Dot14`.


mikeday (approved):
Looks fine to me!

PR #320, opened 2026-08-11
Fix StateError, reimplement `Value::coerce_mapped_value` to drop `Permit` wrapper

archaephyrryx:
Assorted technical debt fixes, mainly focused on two areas:

- Parser: More graceful state-management and error-over-panic refactoring of `doodle::parser::offset`, along with general improvements to `doodle::parser::error::ParseError`/`StateError`; distinguished operational semantics for parser-level errors vs data-level errs w.r.t. `Format::Permit`
- Interpreter: introduce `Coerced<T>` type to make Permit recursion-safe in `coerce_mapped_value` and `extract_mapped_value` without losing the distinction between `Value::Permit(Ok(v))` and `Value::Permit(Err(Some(v)))`; this enables downstream processors (like `doodle::output::tree`) to decide whether they should signal that a value was a custom Permit fallback rather than being data parsed from the buffer directly  

Additionally rewrites some doc-comments, adds NOTE comments in various places, and defines the `Numeric` trait in `prelude` to reify a hypothetical trait hard-coded into the logic of `doodle::numeric::codegen::synthesize_unaryop`.


mikeday (approved):
That looks fine.

